PR #12, opened 2016-03-11
use heroku-buildpack-emberjs

fivetanley:
Uses's @hone's heroku-buildpack-emberjs to deploy instead of our custom bin script.


.buildpacks
```diff
@@ -0,0 +1,2 @@
+https://github.com/heroku/heroku-buildpack-nodejs.git#v89
```
hone: the emberjs also includes node. Is there a reason you're using v89 of node?

package.json
```diff
@@ -11,7 +11,7 @@
     "build": "ember build",
     "start": "ember server",
     "test": "ember test",
-    "postinstall": "bower i && ./bin/heroku-deploy"
+    "postinstall": "bower i"
```
hone: you also shouldn't need to do this.

.buildpacks
fivetanley: Ah, I did not know it also includes node.

package.json
fivetanley: it's nice in development to have npm run bower install.

PR #18, opened 2016-03-25
Update x-select & ember

Robdel12:


fivetanley:
This branch has conflicts with master. Also, we should pin `x-select` to a specific revision instead of `master`.

Issue #5, opened 2016-03-10
Error page


fivetanley:
When an error occurs in the application, we should display a "Whoops! Something went wrong." or similar page

acorncom:
Looks like this was closed by the PR linked in above

PR #19, opened 2016-03-27
[ENHANCEMENT] Adding an error page to handle 404 and other errors

### LINKED ISSUE #4: Needs a 404 state
For example, when you navigate to https://ember-api-docs.herokuapp.com/lol1234, we should display a "Not found" page of some kind

sivakumar-kailasam:
The 404 page is similar to the ember website's content with a link to the official repo while the error content is a generic message. Both flows mentioned here use the same route.

Fixes #4 and #5. Here are some screenshots,
screenshot 2016-03-26 19 21 00
screenshot 2016-03-26 19 21 14

### DIFF
--- a/app/helpers/eq.js
+++ b/app/helpers/eq.js
@@ -1,0 +1,7 @@
+import Ember from 'ember';
+
+export function eq(params) {
+  return params[0] === params[1];
+}
+
+export default Ember.Helper.helper(eq);
--- a/app/styles/components/_all.scss
+++ b/app/styles/components/_all.scss
@@ -6,4 +6,5 @@
 @import "back-to-top";
 @import "highlight";
 @import "toc";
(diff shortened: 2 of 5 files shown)


fivetanley:
This is great! Although I think we could consolidate this into one `error.hbs` template so the URL doesn't change.

sivakumar-kailasam:
@fivetanley I got the changes into the error template, but the build keeps failing since the travis node can't download phantomjs binary. I even reworded my commit just to restart it once :smile:  but its till keeps failing :disappointed: . Can you do a retry on the travis side?

sivakumar-kailasam:
Looks like it was something intermittent between travis and bitbucket. Its all clean now!
(there was some force pushing to redo travis builds :smile: )

fivetanley:
Thanks!

PR #21, opened 2016-03-27
Use Heroku Emberjs Buildpack

hone:
This PR removes the unnecessary components when using the [emberjs heroku buildpack](https://github.com/heroku/emberjs). To setup the buildpack, you need to use the heroku toolbelt `buildpacks` command:

```
$ heroku buildpacks:set https://codon-buildpacks.s3.amazonaws.com/buildpacks/hone/emberjs.tgz
```


fivetanley:
:pray:

PR #29, opened 2016-03-29
Scrolly sidebar

tehviking:
Upsides: Lets the sidebar scroll separately from the main content

Downsides: makes the sidebar fixed width and doesn't scale down to mobile


toddjordan:
@tehviking noice!  Thanks for the contribution :-)

PR #30, opened 2016-03-29
Fastly Support

hone:
This replaces AWS/S3 with Fastly on Heroku. I ran into [a concurrency issue](https://gist.github.com/hone/4db40e4b3a2199c0c1e6) with `express-cluster` in master of `ember-fastboot-server`. Setting `WORKER_COUNT=1` as an env var fixed this.


fivetanley:
I think we ended up merging this manually in a different commit.

PR #32, opened 2016-03-29
Setting up collapsible classes list

### LINKED ISSUE #24: Classes list should expand/unexpand like on the guides site
For example: https://guides.emberjs.com/v2.4.0/

Starts off closed:

When clicked, opens:

On the guides site, there's a neat little animation that I'm guessing is done by using a CSS transition.

This would be a great component! If you can't find something that works off the shelf (**that also works in fastboot**), unit tests would be great. It would also be great if we could make this accessible (cc @Robdel12)

acorncom:
Fixes #24

### DIFF
--- a/app/components/table-of-contents.js
+++ b/app/components/table-of-contents.js
@@ -1,0 +1,9 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  actions: {
+    toggle() {
+      this.$('ol.toc-level-1').slideToggle(200);
+    }
+  }
+});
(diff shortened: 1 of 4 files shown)


acorncom:
Tests still need to be written.  And ideally animations would be nicer than they are right now ...

acorncom:
@fivetanley Looks like this is ready to go

fivetanley:
@acorncom this looks good, can you fix the merge conflicts when you have a chance? sorry for the late response.

acorncom:
@fivetanley Updated here.  Re: late response, no problem, we all needed a bit of a breather post conference ;-)

PR #33, opened 2016-03-30
add loading spinner

gregone:
ember-docs-loading

This PR addresses https://github.com/fivetanley/ember-api-docs/issues/10

This is a first pass at adding a loading spinner. I chose the square-ish animation since it kind of matches the geometric background in the header. That said, if we decide to go for something more rounded instead, I'm fine with changing it...

### DIFF
--- a/app/components/loading-spinner.js
+++ b/app/components/loading-spinner.js
@@ -1,0 +1,5 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  classNames: ['loading-spinner']
+});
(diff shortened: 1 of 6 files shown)


gregone:
@fivetanley Looks like the build errored for non-ember reasons on Travis, and I think I don't have access to the "restart" button. Can you give it a shot?

acorncom:
@gregone I think you'd be able to re-kick the job if you do a `git commit --amend` and then do a `git push --force` back to your branch.  That work for you?

Issue #11, opened 2016-03-11
Remove private classes from sidebar


fivetanley:
To do this, we would need to change the documents generated by https://github.com/fivetanley/ember-jsonapi-docs.

ideally, each `project-version` document would not have a `classes` relationship anymore, but `private-classes` and `public-classes`.

I can handle this one.

Gaurav0:
I've added this to my PR at https://github.com/ember-learn/ember-jsonapi-docs/pull/2

toddjordan:
I think we need to de-emphasize the toggle here.  Right now its bigger/bolder than the heading.

Also, I think we need to identify private classes in the menu when shown.  italics/lighter font maybe?

![image](https://cloud.githubusercontent.com/assets/3609063/20151730/42a14f78-a689-11e6-8b9c-899935ae6e62.png)

Finally, I think what we talked about back in the day was to have an internals subdomain where users interested in internal apis could go to, but not show this to typical Ember developers.  I think we can handle this with an application config attribute. and set it on start, so when page is loaded, its already showing privates...

Gaurav0:
I've moved Show Private to bottom to hide it better and made it smaller font size. Remaining issues are:

1) Create flag for Internals site
2) Hide checkboxes in main site

toddjordan:
going ahead and closing this issue.  Making a new issue with remaining work.

PR #44, opened 2016-05-17
Private toggle for classes

Gaurav0:
Issue #11

### DIFF
--- a/app/components/table-of-contents.js
+++ b/app/components/table-of-contents.js
@@ -1,6 +1,26 @@
 import Ember from 'ember';
 
+const { computed, inject, RSVP } = Ember;
+
 export default Ember.Component.extend({
+  store: inject.service(),
+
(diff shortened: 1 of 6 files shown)


Gaurav0:
Screen Capture:

show private toggle

fivetanley:
I'll have to give this a shot against the production db. I was thinking we could do the public/private filtering on the backend for performance reasons and have public/private be two relationships. What do you think about that @Gaurav0?

Gaurav0:
Absolutely makes sense. How do I get the latest db?

fivetanley:
Sorry for the delay! They are now in the db for all versions here. https://fivetanley.cloudant.com/docs/project-version-ember-2.5.1

fivetanley:
I think we could add a `publicClasses` and `privateClasses` relationship on the `ProjectVersion` model now

Gaurav0:
See PR #48

PR #60, opened 2016-11-03
Make redirect test independent of database

Gaurav0:


tests/acceptance/redirects-test.js
```diff
@@ -7,9 +7,11 @@ moduleForAcceptance('Acceptance | redirects');
 test('visiting /', function(assert) {
   visit('/');
 
-  const last = "2.6.0"; // need help for how to get this
+  return andThen(() => {
+    const store = this.application.__container__.lookup('service:store');
```
josemarluedke: Shouldn't this use the `getOwner` api?

Gaurav0: getOwner(this.application) is undefined.

rwjblue: `this.application.__deprecatedInstance__` is better than `__container__` usage, but still annoying.  This is being addressed in https://github.com/emberjs/rfcs/pull/119, but obviously that doesn't help us today...

PR #71, opened 2016-11-14
Return types + bugfixes + tests

Gaurav0:
Bugfixes include:

- tables showing 'over' header
- Could not navigate from classes to methods

### DIFF
--- a/app/controllers/project-version/class.js
+++ b/app/controllers/project-version/class.js
@@ -1,10 +1,17 @@
 import Ember from 'ember';
 import _ from 'lodash/lodash';
 
-const { computed } = Ember;
+const { computed, inject, Controller } = Ember;
 
+export default Controller.extend({
+  router: inject.service('-routing'),
+  routeName: computed.readOnly('router.currentRouteName'),
+  parentName: computed('routeName', function() {
+    const routeName = this.get('routeName');
+    const routes = routeName.split('.');
+    return routes.slice(0, 2).join('.');
+  }),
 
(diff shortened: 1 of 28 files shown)


app/controllers/project-version/class.js
```diff
@@ -1,10 +1,17 @@
 import Ember from 'ember';
 import _ from 'lodash/lodash';
 
-const { computed } = Ember;
+const { computed, inject, Controller } = Ember;
 
+export default Controller.extend({
+  router: inject.service('-routing'),
+  routeName: computed.readOnly('router.currentRouteName'),
+  parentName: computed('routeName', function() {
```
Gaurav0: While this could be done, it would be more code because it has to be inserted into several routes. While a mixin or inheritance could help, I think this is easier.

toddjordan: yeah, its fairly straight forward.  just feels a little weird to be using the route service  in the controller.  Not necessarily bad, just feels awkward.  I'm fine keeping as is though.

toddjordan:
Just an fyi, I'll also be bumping into routes this week as well I create the tabbed interface.  Looking pushing some of the logic from route templates to components (pr coming soon)  I think its reasonable though to have each as a sub route where we'd be addressing `classes/<class>/methods/<method>`, making them nice and linkable

Gaurav0:
@toddjordan Yes, feel free to refactor. I think having more stuff in components would be good.

Issue #51, opened 2016-10-13
Show main sections above our property, methods section


acorncom:
In the old api viewer, we had this section across the top to allow folks to quickly focus on one area ...
![screen shot 2016-10-13 at 3 31 34 pm](https://cloud.githubusercontent.com/assets/802505/19367853/4f515fb8-915a-11e6-9f46-5049cd010f6b.png)

PR #77, opened 2016-11-18
[WIP] Add tabbed layout

toddjordan:
Addresses #51 

Still need to get acceptance tests back working, and get the query param to work across tabs.


app/services/filter-data.js
```diff
@@ -0,0 +1,8 @@
+import Ember from 'ember';
+
+export default Ember.Service.extend({
```
toddjordan: Using a service to persist filterData across screens (namespace and class), so that selections don't get reset.

app/templates/project-version/class.hbs
```diff
@@ -37,79 +37,95 @@
...
-          </li>
-        {{/each}}
-      </ul>
-    {{else}}
-      No documented events.
-    {{/if}}
-  </section>
-
-  {{outlet}}
+      <section>
```
toddjordan: We want filter checkboxes to show up across tabs, so we now show them here instead of in the class-index template.

toddjordan:
working on adding some more acceptance tests around some of the stuff here.
1. switching between class and namespace and ensuring filter selections persist.  
2. Selecting tabs and ensuring proper urls.
...

PR #78, opened 2016-11-21
Add tabbed layout

Gaurav0:
Addresses #51

Picks up on @toddjordan 's work and changes approach to use a `link-to` instead of a generic tab component.


toddjordan:
merged into original PR. Closing.

Issue #83, opened 2016-11-22
activate tab when switching by clicking item in index


toddjordan:
in current implementation we use the link-to active indicator to highlight selected tabs.  This works fine when clicking tabs, but not when linking from the index (because of individual subroutes) we need to keep the tab active.

To do this we can create a helper that prints each sub route of a tab.  Something like `current-when=(list-routes model (concat parentName ".events")`

toddjordan:
No need for helper.  setting current-when to the parent route will activate on sub-routes.

PR #90, opened 2016-11-26
Keep tab routes active on link

toddjordan:
Addresses #83 

turns out setting current-when on the routes, will keep link active when a sub-route is in effect.


Gaurav0:
LGTM

Issue #93, opened 2016-11-30
Clicking on certain classes hides the left nav menu


toddjordan:
To reproduce, click on Ember.ApplicationInstance, Ember.Controller, or Ember.CoreObject

PR #101, opened 2016-12-04
Fix layout css for larger screens

toddjordan:
Simplified things a bit. took z-index out of the equation

Addresses #93


Gaurav0:
IMO, this caused the layout to be too wide on larger screens. Creating issue.

Issue #85, opened 2016-11-22
hide private toggles for public deployment


toddjordan:
In issue #11 we have the user the ability to toggle off/on display of private classes.  The plan currently is to default the public api site to have private off, and possibly hide the toggle.  At the same time we want to have an interals site where everything is on by default and toggle-able.

toddjordan:
Moving to Feature Board https://github.com/ember-learn/ember-api-docs/projects/5

PR #116, opened 2016-12-07
make show private / deprecated checkbox fixed to bottom on large screens

MartinMalinda:


Gaurav0:
@toddjordan @fivetanley @locks @acorncom Please chime in.

I'm not sure we want to do this. The idea of placing it on the bottom was to hide it somewhat. See issue #85

MartinMalinda:
Right, it was just an idea. It partially solves a problem when clicking the checkbox makes it move as new content is being added. It depends on how often will this checkbox be clicked. If it's rarely used it's probably wise to hide in the bottom.

toddjordan:
Going to go ahead and close this one.  We'll have it if we want to reconstitute it in the future.

PR #117, opened 2016-12-07
add mixin for resetting scroll position on transition

MartinMalinda:
So far probably not ready to merge. I'll add acceptance test tomorrow.

Scroll position is reset only four route path changes <2 levels deep, so position is not reseted on clicking Index/Method tabs.


Gaurav0:
Please see ESLint issues.

app/router.js
```diff
@@ -1,9 +1,11 @@
 import Ember from 'ember';
 import config from './config/environment';
 
+const {$} = Ember;
```
Gaurav0: Where is this being used?

app/router.js
```diff
@@ -1,9 +1,11 @@
...
+
 const Router = Ember.Router.extend({
   location: config.locationType,
-  rootURL: config.rootURL
+  rootURL: config.rootURL,
```
Gaurav0: Why add comma?

MartinMalinda:
Should have ran all tests before push. It should be fine by now.

MartinMalinda:
The first assert fails in firefox. Let me fix that. Should I squash commits?

toddjordan:
@MartinMalinda I should be able to squash on merge, when fixed.  Thanks!

MartinMalinda:
The test passes locally on all browsers for me. So it must be some issue with a headless firefox or maybe xvfb.

I'm not really sure how to debug this.

Gaurav0:
If you change `test` to `skip` and create a github issue, I'll accept.

toddjordan:
Downloaded branch locally and un-skipped the test.  Fails on phantom, chrome, and firefox for me :-(

MartinMalinda:
If it fails also in PhantomJS then maybe it's not just caused by screen size :/
I'll try to come up with something smarter. Maybe manually setting the article in overflow:scroll mode and increase it's height so it's always scrollable regardless on screen size and other issues.

toddjordan:
Also tried it manually and didn't work:
api-scrolling-notfixed

MartinMalinda:
Ugh. It definitely works for me. I'll see what I can do.

app/mixins/reset-scroll-position.js
```diff
@@ -0,0 +1,24 @@
...
+    const currentThirdParamName = Object.keys(currentRouteParams)[2];
+    const newThirdParamName = Object.keys(newParams)[2];
+
+    return currentRouteParams[currentThirdParamName] === newParams[newThirdParamName];
+  },
+
+  actions: {
+    willTransition(transition) {
+      if (!this._hasSameThirdParam(transition)) {
+        $('main > article').scrollTop(0);
```
toddjordan: the dom has changed last couple of weeks.  Instead of `$('main > article')`, you'll need `$('main' > 'section')`, or even better, `$(.container > .content)`

toddjordan: The scroll happens early, popping the scroll to the top before the actual transition.  it would be nice if we could find a way to scroll when it actually re-renders.

app/mixins/reset-scroll-position.js
```diff
@@ -0,0 +1,24 @@
+import Ember from 'ember';
+
+const {$} = Ember;
+
+export default Ember.Mixin.create({
+
+  _hasSameThirdParam(transition) {
```
toddjordan: What is this checking?

MartinMalinda: It is checking old and new third param value. So if it's 'debug' === 'debug', scroll reset won't happen. That way there's no jump when switching tabs (methods, properties, etc).

app/mixins/reset-scroll-position.js
MartinMalinda: *facepalm* that must be the reason it's failing. My bad. I'll pull into my branch.

MartinMalinda: `didTransition` would be perfect and this won't happen but it does not receive the `transition` object. Maybe I could `lookup` the router and get sufficient data from there but that's a little bit dirty.

MartinMalinda:
The acceptance test is passing now.

PR #120, opened 2016-12-10
add mixin for resetting scroll position on transition (2)

### LINKED ISSUE #105: Reset scroll to top when navigating routes
When you scroll down a longer item, and then switch to another item that requires scrolling the new route loads partially scrolled down-screen.  We should reset scroll to top on route change.

MartinMalinda:
Rebased and squashed version of #117.

Should fix #105.


app/mixins/reset-scroll-position.js
```diff
@@ -0,0 +1,32 @@
...
+    const currentThirdParamName = Object.keys(currentRouteParams)[2];
+    const newThirdParamName = Object.keys(newParams)[2];
+
+    return currentRouteParams[currentThirdParamName] === newParams[newThirdParamName];
+  },
+
+  actions: {
+    willTransition(transition) {
+      if (!this._isChangingTab(transition)) {
+        this.set('routingService._shouldResetScroll', true);
```
toddjordan: I think you can set shouldResetScroll on this object and not on the routing service.  I think we are guaranteed to have the same route instance between will and did transition.

MartinMalinda: I tried that but one tests fails with that solution: `scroll position is resetted after transition: project-version.class.index to project-version.module.index.` The mixin is also placed at route `class` and route `module` so `didTransition` and `willTransition` can be called from different places.

toddjordan: On your second point... the mixin is not a single instance it gets mixed in to currently instantiated route objects, so the fact that multiple routes use it is inconsequential.  the state here only matters for the current route.

I think I'm ok with you using a service.  A little queazy about using the -routing service though.  We could create a route-scroll-reset-service, and push the state there, and even the logic.

also, if we are accessing _shouldResetScroll externally from the service it resides in, then the property is not private and should not prepend with `_`.  Though if we were to push the logic into a service, then we could access it as a private.

MartinMalinda: Agreed. This is a good idea. I got rid of the mixin alogether. Now it's needed to define actions and the transition hooks on all places, but at least it's explicit and more logic can be easily added there in the future if needed.

toddjordan:
❤️  Thanks @MartinMalinda

Issue #102, opened 2016-12-04
Should modules come before namespaces?


toddjordan:
in the left nav of the current api docs, modules are shown before namespaces.  In the new api docs, they are shown after.  Was there a reason for this?  It may cause some confusion for those used to it they other way.  (for example there's an `ember` in both, so one might click on ember module, thinking it was a namespace.)

toddjordan:
I'm inclined to go ahead and change the order if no objections...  cc @Gaurav0

Gaurav0:
I don't care so go ahead. :)

acorncom:
I think showing modules before hand is better UI 👍

toddjordan:
So leave as is?

PR #127, opened 2016-12-22
modules before namespaces

toddjordan:
addresses #102 

switches the order of modules and namespaces to be consistent with the current api layout.


Gaurav0:
LGTM

Issue #96, opened 2016-12-01
Namespace/Module Incosistencies


Gaurav0:
References to namespaces that don't exist and other data issues break the app.

toddjordan:
@Gaurav0 can you be specific about the "other data issues"?

Gaurav0:
@toddjordan So far, the only "other" data issues I have directly found are references to private methods and properties from public ones.

However, the app should never break regardless of any data issue. A reference to a namespace, module, class, property, method, or event that either doesn't exist or isn't currently visible according to toggles should result in a friendly error message.

toddjordan:
ok, in 2.10, I clicked on every namespace and module.  The only one that doesn't come up is the 'support' module.  in 2.9 the glimmer module shows headings but no data

toddjordan:
Ok, I'm going to use this issue to highlight specific differences in content between the new api-docs and the currently deployed version.

toddjordan:
- [ ] ember-* is missing from each module (except ember)
- [ ] ember-extension-support is "support" in the new app (and loads blank)

new: 
![image](https://cloud.githubusercontent.com/assets/3609063/20868868/90180a76-ba34-11e6-88de-ab5dfdbba8c4.png)

old:
![image](https://cloud.githubusercontent.com/assets/3609063/20868867/823e16b6-ba34-11e6-99bc-7cc8776562fb.png)

- [ ] `Ember.ApplicationInstance` and `Ember.Test` load as Namespaces in other namespace details when in actuality they are Classes

![image](https://cloud.githubusercontent.com/assets/3609063/20868915/7c6360b0-ba35-11e6-87f4-f5d64ea721d3.png)

toddjordan:
oddly, things are showing up as namespaces in the data that don't seem to be namespaces:
![image](https://cloud.githubusercontent.com/assets/3609063/21156375/9b993d5a-c143-11e6-8f04-17d0ba49ec2e.png)

toddjordan:
In addition, the current api docs at emberjs.com/api combine "Classes and Namespaces" for modules, and it is unclear to me atm how our docs generation code differentiates.

toddjordan:
Ok, I think I understand somewhat what's happening.

1. modules in the left nav are defined from source code with `@module` or `@submodule`
2. the classes within a module are defined from any `@class` defined within a source file with the `@module` or `@submodule` label
[...]

PR #130, opened 2017-02-03
Consolidate namespace routes to the classes route

toddjordan:
Addresses #96 by making the linking behavior the same as it currently is at https://emberjs.com/api

In a submodule, classes and namespaces will now show up in a common, sorted section called "Classes and Namespaces" (as opposed to separate sections as previously)

image

Current App:
image

_(in our new version its at least sorted)_

All classes and namespaces will be addressable under /ember/<version>/classes/<class name>, both from submodule details, and from the left nav.

To support separation of classes and namespaces, we'll need to do some work on how namespaces and classes are identified within a submodule in the yuidoc, likely requiring changes to the ember code documentation.  This is a bit bigger issue than what we want to tackle at this point, so the decision was made in the last learn team meeting to go for the same routing behavior as what is already there, that being all classes and namespaces routable under 'classes'.

### DIFF
--- a/app/controllers/project-version.js
+++ b/app/controllers/project-version.js
@@ -29,7 +29,7 @@ export default Ember.Controller.extend({
   }),
 
   getModuleRelationships(versionId, moduleType) {
-    let relations = this.getRelations('public-modules');
+    let relations = this.getRelations(moduleType);
     return relations.map(id => id.substring(versionId.length + 1))
   },
 
--- a/app/controllers/project-version/module.js
+++ b/app/controllers/project-version/module.js
@@ -1,5 +1,6 @@
 import Ember from 'ember';
 import ClassController from './class';
+import _ from 'lodash/lodash';
 
 const { computed } = Ember;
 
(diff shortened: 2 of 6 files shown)


Gaurav0:
I really don't like having "Classes and Namespaces". I thought we discussed instead having /namespace/* redirect to /class/* if it doesn't find a namespace but does find a class?

toddjordan:
The problem goes beyond the url.  Ember right now can't accurately differentiate between a class and a namespace in regard to their relationship to a submodule.  The issue has some explanation.

"Classes and Namespaces" are what the api has currently.  Not saying we'll never make this better. Just saying that to make it better will require a bigger effort than what we want to spend to get this out the door for an MVP.  I can create an issue and assign it for the next iteration of this app.

Gaurav0:
Am I missing some of your changed files? I pulled your branch and don't see "Classes and Namespaces" section

toddjordan:
did something get cached?  Notice in the Changed files that we updated module.hbs.  

In the UI, just click on one of the modules, such as `ember-glitter` on the left nav.

Gaurav0:
Ok. Misunderstood. I thought the side navbar would show classes and namespaces.

LGTM.

PR #136, opened 2017-02-21
Upgrade to Ember 2.11.1

toddjordan:
#136


package.json
```diff
@@ -52,26 +52,28 @@
...
     "ember-export-application-global": "^1.0.5",
     "ember-font-awesome": "2.2.0",
     "ember-inflector": "^1.9.4",
-    "ember-load-initializers": "^0.5.1",
+    "ember-load-initializers": "^0.6.0",
     "ember-lodash": "0.0.11",
     "ember-pouch": "github:gaurav0/ember-pouch#c89600c7b8c14c919b9b808a4ea6d9d6666ca047",
     "ember-resolver": "^2.0.3",
     "ember-route-action-helper": "2.0.0",
+    "ember-source": "~2.11.0",
```
Gaurav0: Can we make this 2.11.1?

tests/helpers/start-app.js
```diff
@@ -5,8 +5,8 @@ import config from '../../config/environment';
 export default function startApp(attrs) {
   let application;
 
-  // use defaults, but you can override
-  let attributes = Ember.assign({}, config.APP, attrs);
+  let attributes = Ember.merge({}, config.APP);
+  attributes = Ember.merge(attributes, attrs); // use defaults, but you can override;
```
Gaurav0: This revert is only needed in addons for compat with Ember 2.4. As this is an app, lets keep the assign code.

package.json
locks: 2.11.2 is out with the missing fixtures fix, fyi

toddjordan: I'll up the version 👍

toddjordan: though it should be picking up patch releases with ~

package.json
```diff
@@ -52,26 +52,28 @@
...
     "ember-cli-inject-live-reload": "^1.4.1",
     "ember-cli-qunit": "^3.0.1",
     "ember-cli-release": "^0.2.9",
     "ember-cli-sass": "^5.6.0",
+    "ember-cli-shims": "^1.0.2",
     "ember-cli-test-loader": "^1.1.0",
     "ember-cli-uglify": "^1.2.0",
     "ember-composable-helpers": "^0.20.0",
-    "ember-data": "~2.10.0",
+    "ember-data": "^2.11.0",
```
Gaurav0: I think we should stick with ~ here.

package.json
```diff
@@ -40,7 +40,7 @@
     "ember-ajax": "^2.4.1",
     "ember-anchor": "~0.1.8",
     "ember-browserify": "~1.1.13",
-    "ember-cli": "2.10.0",
+    "ember-cli": "^2.11.1",
```
Gaurav0: No ^ here. We want ember-cli updates to take place manually.

package.json
toddjordan: Got this from here: https://github.com/ember-cli/ember-new-output/compare/v2.10.1...v2.11.1

toddjordan: Maybe because ember-data is more semver friendly?

package.json
toddjordan: yeah, originally I had ~ which was fine.  Didn't really even need to make it 2.11.1 (because the ~).  Will switch back

Issue #104, opened 2016-12-04
Modules should show parent link


toddjordan:
Modules show a link to the parent module (usually ember)

![image](https://cloud.githubusercontent.com/assets/3609063/20868998/1c3f8932-ba37-11e6-8547-887329d1b31e.png)

MartinMalinda:
It is fetched from `@module`
https://github.com/emberjs/website/blob/master/source/api/module.html.erb#L6

Which is provided by YUIDoc output:
https://github.com/emberjs/website/blob/master/data/api.yml#L1635

ember-jsonapi-docs do not expose this value yet:
![screen shot 2016-12-09 at 18 34 11](https://cloud.githubusercontent.com/assets/9468793/21058453/4de48f86-be3e-11e6-96df-d78ac31d3cf3.png)

Some `parentModule` relationship will probably be needed.

toddjordan:
@esantelises is currently looking at this issue

acorncom:
Closed by #139

PR #139, opened 2017-02-26
Add parent link to modules

esantelises:
Addresses #104, display submodule parent.

Depends on https://github.com/ember-learn/ember-jsonapi-docs/pull/20

image

### DIFF
--- a/app/models/module.js
+++ b/app/models/module.js
@@ -6,5 +6,6 @@ const { attr } = DS;
 export default ClassModel.extend({
   submodules: attr(),
   classes: attr(),
-  namespaces: attr()
+  namespaces: attr(),
+  parent: attr()
 });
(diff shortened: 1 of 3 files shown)


Gaurav0:
Can you look at the test failure?

esantelises:
@Gaurav0 since acceptance tests depends on some data to be loaded for some versions like `1.0.0`, my 
guess is `parent` attribute is still missing on CI data not sure if after https://github.com/ember-learn/ember-jsonapi-docs/pull/20 is merged i would just need to re-run the build or if need to update the data somewhere else? (all test are passing locally)

toddjordan:
Fix works on my local env as well.  I've merged the jsonapi-doc side of the fix, se we need to deploy/refresh the cloudant db.

Fix LGTM.  Thanks!

Issue #140, opened 2017-03-03
Remove Bourbon Deprecations


toddjordan:
Currently seeing this when starting the app: 

```
WARNING: [Bourbon] [Deprecation] `em` is deprecated and will be removed in 5.0.0.
Backtrace:
	tmp/sass_compiler-input_base_path-MuVZEZ9A.tmp/0/app/styles/functions/_px-to-em.scss:7, in function `em`
	tmp/sass_compiler-input_base_path-MuVZEZ9A.tmp/0/app/styles/base/_grid-settings.scss:10

WARNING: [Bourbon] [Deprecation] `em` is deprecated and will be removed in 5.0.0.
Backtrace:
	tmp/sass_compiler-input_base_path-MuVZEZ9A.tmp/0/app/styles/functions/_px-to-em.scss:7, in function `em`
	tmp/sass_compiler-input_base_path-MuVZEZ9A.tmp/0/app/styles/base/_grid-settings.scss:11

WARNING: [Bourbon] [Deprecation] `appearance` is deprecated and will be removed in 5.0.0. We suggest using an automated prefixing tool, like Autoprefixer.
Backtrace:
	tmp/sass_compiler-input_base_path-MuVZEZ9A.tmp/0/app/styles/_bourbon-deprecate.scss:10, in mixin `-bourbon-deprecate`
	tmp/sass_compiler-input_base_path-MuVZEZ9A.tmp/0/app/styles/_bourbon-deprecate.scss:17, in mixin `-bourbon-deprecate-for-prefixing`
	tmp/sass_compiler-input_base_path-MuVZEZ9A.tmp/0/app/styles/css3/_appearance.scss:2, in mixin `appearance`
	tmp/sass_compiler-input_base_path-MuVZEZ9A.tmp/0/app/styles/base/_buttons.scss:2

WARNING: [Bourbon] [Deprecation] `appearance` is deprecated and will be removed in 5.0.0. We suggest using an automated prefixing tool, like Autoprefixer.
Backtrace:
	tmp/sass_compiler-input_base_path-MuVZEZ9A.tmp/0/app/styles/_bourbon-deprecate.scss:10, in mixin `-bourbon-deprecate`
	tmp/sass_compiler-input_base_path-MuVZEZ9A.tmp/0/app/styles/_bourbon-deprecate.scss:17, in mixin `-bourbon-deprecate-for-prefixing`
[...]

PR #142, opened 2017-03-09
Fix Issue #140: Remove bourbon deprecations

steveszc:
Fix Issue #140 
- Removed/replaced deprecated Bourbon features
- Implemented autoprefixer to replace browser prefixing previously provided by deprecated Bourbon features


toddjordan:
This is great.  Tried it out, no more deprecations and it looks good.  Thanks so much for helping out!

Issue #134, opened 2017-02-17
Store Application assets using Service Workers


toddjordan:
Part of #132

Application Data can be stored using service worker/app cache, so that its available for offline browsing.

Here's the planned approach using service workers:

App: 
- User visits a route (ex /ember/v2.11.2/classes/Ember.Component.html)
- App fetches a manifest json file for a given version of the project (ember v2.11.2, etc)
  The Manifest contains the names/urls of all the items in the project for that particular version.
- Also fetches the doc file (json) the item the route is pointed at (for example Ember.Component)

Service Worker:
- registered when the app loads in the browser
- listens for the fetch of the manifest.  
-- attempts to load it from cache if there, otherwise fetches from the server.
-- returns the result from either the cache or the server as the response.
-- kicks off a process that goes through the list of items, fetching each from its url to hydrate the cache.
- listens for the fetch of a doc file
-- Each individual item fetched by the app (classes, modules, etc) are cache first, falling back to a server fetch.

toddjordan:
open question would be whether to use the ember-service-worker architecture, and write a plugin, or roll our own registration code.   With the addon, we can get some stuff for "free" (registration, caching of index, app files).  Am doing a prototype branch atm.

toddjordan:
going forward with marten's plugin architecture.    Have a pr that installs it and includes the index and basic asset caching.  Also started a script to pull down docs data in the background (waiting on manifest)

toddjordan:
Blocked by #214

toddjordan:
Moving to feature board https://github.com/ember-learn/ember-api-docs/projects/5

PR #144, opened 2017-03-12
Add service worker

toddjordan:
This is an initial implementation of service workers, simply using the ember-service-worker plugin. 

Work is ongoing surrounding background loading of data assets

Partially addresses #134


toddjordan:
#134

PR #152, opened 2017-03-16
remove font awesome and add fa-pencil inline svg

### LINKED ISSUE #150: font awesome is loaded just to show one pencil icon
It's 7kb of css and 76kbs of fonts. I think inline SVG should be better, unless there's a plan to have more icons in the future.

MartinMalinda:
fixes #150


package.json
```diff
@@ -65,7 +65,6 @@
     "ember-data": "^2.11.0",
     "ember-disable-proxy-controllers": "^1.0.1",
     "ember-export-application-global": "^1.0.5",
-    "ember-font-awesome": "2.2.0",
```
sivakumar-kailasam: @MartinMalinda can you remove the dependency from the yarn.lock too? There's a corresponding bower dependency as well.

sivakumar-kailasam:
There are some fontello related files in styles folder that can be removed with this PR since they aren't being used.

MartinMalinda:
I have also removed select2 styles while I was at it. Maybe the swiftype css could be removed as well?

toddjordan:
Also unsure why we are pinned to a github commit of select2 as opposed to semantic versions.  That can be part of a separate pr updating dependencies though.  Pulling this down now to check it out...

sivakumar-kailasam:
Speaking of select, api-docs uses emberx-select & guides uses power select (think so). Should we move to power select as well?

app/templates/project-version/class.hbs
```diff
@@ -1,5 +1,5 @@
 <article class="chapter">
-  <a class="heading__link__edit" href="{{github-link model.project.id model.projectVersion.version model.file model.line}}" target="_blank" rel="noopener">{{fa-icon "pencil"}}</a>
+  <a class="heading__link__edit" href="{{github-link model.project.id model.projectVersion.version model.file model.line}}" target="_blank" rel="noopener">{{partial "svg/fa-pencil"}}</a>
```
MartinMalinda: I can add a simple one via CSS pseudo element, reading from some data-title attr.

toddjordan: i'll go ahead and merge.  It can be done in a separate pr

PR #154, opened 2017-03-16
Mobile layout adjustments

MartinMalinda:
This removes extra paddings and margins on mobile and sometimes medium resolutions to fit more content screen.

Also, some elements are wrapped beneath each other.

Grey border for "tabbed layout" has been removed to prevent double paddings and to have more space.

There was also an issue when Events, Properties etc. did not fit into the 320px mobile screen.

screen shot 2017-03-16 at 18 22 03
screen shot 2017-03-16 at 18 21 50

### DIFF
--- a/app/styles/_class.scss
+++ b/app/styles/_class.scss
@@ -24,6 +24,7 @@ article.chapter {
 
 .access-checkbox {
   display: inline;
+  white-space: nowrap;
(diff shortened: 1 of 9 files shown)


toddjordan:
This is great @MartinMalinda thanks!

PR #156, opened 2017-03-17
Add queryParam support for show entities

### LINKED ISSUE #148: Capture show entity from api viewer in query params
When we select to see inherited, private, protected & deprecated entities, they are not captured in URLs as query params. This makes it hard to see the same information when sharing links.

sukima:
Adds a mixin that will translate the filterData properties into a comma separated queryParam (show=). At the moment it is used by the project-version controller since it is the parent of all the controllers that would need the filterData. It is started as a mixin because I thought it had to be part of each controller but (Ember being cool) it can also work in one spot in the parent hierarchy. I left it as a mixin simple because the project-version controller is almost 100 lines large so far.

I aded one test to see the query params show up when you click the check boxes. I also had to fix some tests that had to brittle expectations of the URL. I simply stripped out the query params from the URL since the tests were not asserting anything regarding params.

Closes #148

### DIFF
--- a/app/controllers/project-version.js
+++ b/app/controllers/project-version.js
@@ -2,10 +2,11 @@ import Ember from 'ember';
 import _ from 'lodash/lodash';
 import semverCompare from 'npm:semver-compare';
 import getMinorVersion from "../utils/get-minor-version";
+import FilterParams from '../mixins/filter-params';
 
 const { Controller, computed, A, inject } = Ember;
 
-export default Controller.extend({
+export default Controller.extend(FilterParams, {
 
   filterData: inject.service(),
   showPrivateClasses: computed.alias('filterData.sideNav.showPrivate'),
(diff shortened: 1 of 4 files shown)


sukima:
Tests officially broken on devops problem not code changes:

```console
$ yarn run sync-docs
yarn run v0.21.3
$ s3-cli sync s3://sk-ed/json-docs public/json-docs 
This utility needs a config file formatted the same as for s3cmd
or AWS_SECRET_KEY and AWS_ACCESS_KEY environment variables.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Proof:

```console
1..148
# tests 148
# pass  148
# skip  0
# fail  0
# ok

The command "yarn test" exited with 0.
Done. Your build exited with 1.
```

PR #160, opened 2017-03-17
add css tooltip to github link

MartinMalinda:
Adds a tooltip via CSS that shows up on hover. Since it's hover, it's a solution for desktop only.

screen shot 2017-03-17 at 21 52 22

### DIFF
--- a/app/styles/_class.scss
+++ b/app/styles/_class.scss
@@ -3,6 +3,54 @@ article.chapter {
     float:right;
     width: 24px;
     height: 24px;
+    display: block;
+    position: relative;
+
+    @media (min-width: $mobile-portrait-screen + 1){
+      &::before, &::after {
+        transition: 0.2s opacity, 0.2s transform;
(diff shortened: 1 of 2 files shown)


toddjordan:
Thanks Martin!  Just a few questions.

What was your reason for using this over title/alt attributes?  Was it for the styling capability?  Also, have you tested this cross browser?

MartinMalinda:
Yes, it looks better and it shows up instantly. Accessibility should not be an issue as well. Tested on Chrome, Firefox, Safari. Not Edge/IE though.

sivakumar-kailasam:
We should probably add [ember-a11y-testing](https://github.com/ember-a11y/ember-a11y-testing/) to this app.

toddjordan:
The tooltip should read more like, "View Source"

toddjordan:
@sivakumar-kailasam true.  I'll add an issue

PR #162, opened 2017-03-17
change element of .attributes to div

### LINKED ISSUE #158: .attribute elements render outside of .attributes parent in server side render
Server side render (incorrect)
screen shot 2017-03-17 at 21 10 54
Client side render (correct)
screen shot 2017-03-17 at 21 10 43

This causes a period of incorrect styling, where attributes are not uppercase and bold.
I'll try to look into this later today.

MartinMalinda:
fixes #158

Strangely enough, the server side rendering works correctly for DIV, but not for P element.


toddjordan:
strange...

Issue #166, opened 2017-03-18
Add ember-a11y-testing to the app


toddjordan:
We want to get on the road to making this accessible.  A good start would be to add the tool https://github.com/ember-a11y/ember-a11y-testing to lint our project for spots to focus on.

So to complete this issue, we would want to install the tool and fix the issues.

PR #168, opened 2017-03-19
Adding a11y testing addon & partial fixes

sivakumar-kailasam:
Addresses #166. There are two pending issue categories. One for the select control not having a label & another on color contrast. I thought of replacing x-select with power-select for consistency with the guides. But wanted to get a confirmation here.

Our color combination for links [fails contrast checks](http://webaim.org/resources/contrastchecker/?bcolor=fffdf9&fcolor=dd6a58). The ember-a11y team has been trying to organize efforts on this via https://github.com/ember-a11y/issues/issues/1. So we might need to decide on those with their help. 

//cc @acorncom - since you were already involved on this conversation 🙂

### DIFF
--- a/app/styles/base/_variables.scss
+++ b/app/styles/base/_variables.scss
@@ -28,7 +28,7 @@ $dark-gray: #444545;
 $tan: #fffdf9;
 $white: #fff;
 $black: #000;
-$creme: #FFFBF5;
+$creme: #FDFDFD;
 $linen: #f9e7e4;
 
 $base-background-color: $creme;
--- a/README.md
+++ b/README.md
@@ -2,6 +2,12 @@
 
 This app is used to produce versioned Ember api docs.  To view the data generator producing its data, please visit https://github.com/ember-learn/ember-jsonapi-docs
 
+
 ## Development setup 
 
 Download either the compact/full docs from [ember-jsonapi-docs:releases](https://github.com/ember-learn/ember-jsonapi-docs/releases) and move the json-docs folder inside the  `public` folder.
+
+
+## a11y testing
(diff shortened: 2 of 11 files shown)


app/styles/base/_variables.scss
```diff
@@ -29,7 +29,7 @@ $dark-gray: #444545;
 $tan: #fffdf9;
 $white: #fff;
 $black: #000;
-$creme: #FFFBF5;
+$creme: #FDFDFD;
```
toddjordan: Not a color guy, but doesn't this change the color of the background of the entire app?   Maybe we should have a ux designer look at our palette and suggest something with proper contrast.  In fact I think people have looked at this for the website.  

@locks Has the color palette been part of the website design updates that have happened lately?

sivakumar-kailasam: Oh sorry for not mentioning this, but this particular color was taken off the guides 😄

toddjordan: oh, well, that'll work then.  I'm pulling the updates and running locally.  We'll see how it looks :-)

toddjordan:
This is cool, but maybe we should add the config section to environment.js and have audit off by default.  I think the screen display is going to troll new contributors (it suprised me :-P )

And then make a followup issue that would have someone turn it on, clean up all existing issues, and then leave it enabled.

sivakumar-kailasam:
Sure, adding that.

toddjordan:
I resolved conflicts and merged with master, but the pr didn't like what I did for some reason.  Therefore I'm going to close the PR, since the code has been merged.

PR #170, opened 2017-03-20
update ember and other dependencies

### LINKED ISSUE #164: Fix up dependencies
Several of the npm dependencies are "pinned", which means we are not picking up minor non-breaking updates, which could potentially cause us some grief.

We should try to get our dependencies referencing `^` versions (other than ember-cli and ember-source).

We are on 0.* version of ember-lodash, which is currently at 4.*
We are on 0.* version of ember-truth-helpers, which is currently at 1.0.0

Also, we are dependent on a specific commit of ember-select2.  We should at the current release instead.

MartinMalinda:
fixes #164 and #165


package.json
```diff
@@ -62,19 +62,19 @@
...
-    "ember-lodash": "0.0.11",
-    "ember-network": "0.3.1",
+    "ember-lodash": "^4.17.1",
+    "ember-network": "^0.3.1",
     "ember-resolver": "^2.0.3",
     "ember-route-action-helper": "^2.0.2",
-    "ember-source": "~2.11.2",
-    "ember-truth-helpers": "1.2.0",
-    "emberx-select": "github:thefrontside/emberx-select#05f0c947d321ca59e58cf1213a9552b888ca0201",
+    "ember-source": "2.12.0",
```
toddjordan: I think we should be ok picking up ember patch releases with `~`.

toddjordan:
awesome ❤️  just one code comment.

PR #191, opened 2017-03-24
Set root url for service worker in prod

toddjordan:
so when we are deployed at emberjs.com/api-new/, we are looking for sw-registration.js at the root (emberjs.com/).  We should look instead at /api-new/

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -72,6 +72,7 @@ module.exports = function(environment) {
     ENV.routerRootURL = '/api-new/';
 
   }
+
   ENV.manifest = {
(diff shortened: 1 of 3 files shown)


config/environment.js
```diff
@@ -70,6 +70,9 @@ module.exports = function(environment) {
      * solved for that
      */
     ENV.routerRootURL = '/api-new/';
+    ENV['ember-service-worker'] = {
+      rootURL: '/api-new/'
```
toddjordan: eventually we want to grab these from env variables I'd imagine.

PR #193, opened 2017-03-24
remove broccoli-manifest, add ember-cli-deploy-manifest

toddjordan:


sivakumar-kailasam:
@toddjordan reverted with #195.

PR #194, opened 2017-03-24
add case for switching version on module and namespace route

### LINKED ISSUE #190: Changing Ember versions when on a sub-page resets your location back to project root
If I'm viewing `/ember/2.12.0/modules/ember-metal` and then change my Ember version to 2.8, I would like to end up at `/ember/2.8.3/modules/ember-metal`.  Instead, I'm currently sent to `/ember/2.8.3/namespaces/Ember`

I'm seeing this behavior on our staging system (viewable through http://www.emberjs.com/api-new/)

MartinMalinda:
fixes #190

### DIFF
--- a/app/routes/project-version.js
+++ b/app/routes/project-version.js
@@ -40,18 +40,26 @@ export default Ember.Route.extend({
       let endingRoute, routeName;
 
       switch (routeName = this.router.currentRouteName) {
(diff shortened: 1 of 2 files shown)


sivakumar-kailasam:
Also there seem to be eslint failures

PR #201, opened 2017-03-25
Fix manifest to exclude missing files

sivakumar-kailasam:
Before:
screenshot 2017-03-25 15 08 02

After:
screenshot 2017-03-25 15 10 33

@toddjordan chrome was trying to load some of those missing files so switching to just mention that parent folder name solved this


toddjordan:
Thanks. Looking at that when I fell asleep😛

PR #206, opened 2017-03-26
call trackPage on transitions with requestIdleCallback

### LINKED ISSUE #183: Add Google Analytics tracking to the site
I'd like to get our Google Analytics tracking code added to the new app (`UA-27675533-1`) so that we can track page views within the app over time.  We'd preferably also disable tracking when we're doing development and only track this in production ...

MartinMalinda:
fixes #183

I know it's not MVP but I did it recently in a different app so it was a quick fix for me. It is wrapped in requestIdleCallback so the analytics should not delay time to interactivity.

The acceptance test failed randomly once for me, probably because of concurrency, so I added some extra guards (the `run` call and last `andThen`) and It seems to be fixed. I re-run it like 20 times and it did not fail again.

### DIFF
--- a/app/services/analytics.js
+++ b/app/services/analytics.js
@@ -1,0 +1,28 @@
+import Ember from 'ember';
+import config from 'ember-api-docs/config/environment';
+import { requestIdlePromise } from 'ember-api-docs/utils/request-idle-callback';
+
+const { computed, $ } = Ember;
+
+export default Ember.Service.extend({
+
+  googleAnalytics: computed(function() {
+    return requestIdlePromise({timeout: 1000}).then(() => {
+      return $.getScript('https://www.google-analytics.com/analytics.js');
+    }).then(() => {
+      window.ga('create', config.gaTrackingId, 'auto');
+    });
+  }),
+
+  trackPage(page, title) {
+    return this.get('googleAnalytics').then(() => {
+      return requestIdlePromise({timeout: 1000});
+    }).then(() => {
+      window.ga('send', {
+        hitType: 'pageview',
+        page,
+        title
+      });
+    });
+  }
+});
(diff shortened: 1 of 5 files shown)


app/utils/request-idle-callback.js
```diff
@@ -0,0 +1,17 @@
+import Ember from 'ember';
+
+const { RSVP, run } = Ember;
+
+export function requestIdleCb(cb, opts) {
```
sivakumar-kailasam: Lets not export the callback version.

app/services/analytics.js
```diff
@@ -0,0 +1,28 @@
...
+    }).then(() => {
+      window.ga('create', config.gaTrackingId, 'auto');
+    });
+  }),
+
+  trackPage(page, title) {
+    return this.get('googleAnalytics').then(() => {
+      return requestIdlePromise({timeout: 1000});
+    }).then(() => {
+      window.ga('send', {
```
sivakumar-kailasam: @acorncom is there any other param that can be added to differentiate api-docs from the main site or will the tracking id unique to the api docs?

acorncom: @sivakumar-kailasam for now let's use the tracking ID I provided. It's the same as the main website tracking ID but we should be able to differentiate where this is being used based on paths

app/utils/request-idle-callback.js
```diff
@@ -0,0 +1,17 @@
...
+  if ('requestIdleCallback' in window) {
+    return requestIdleCallback(cb, opts);
+  } else {
+    return run.shceduleOnce('afterRender', cb);
```
acorncom: Looks like a spelling mistake here?

MartinMalinda: Definitely. Thanks for the catch!

PR #209, opened 2017-03-27
Handle /release requests

### LINKED ISSUE #185: Add a /current route that we can use in various places
At times we need the ability to link to "the most current" version of api docs (see https://github.com/ember-learn/ember-api-docs/issues/184 for an example). We need to add a `/current` route to this Ember app that will display the most current version of the api docs.

MartinMalinda:
should fix #185

- [x] get latest project version when visiting /release
- [x] add unit and/or acceptance test
- [x]  make sure that links on /release also link to /release (for SEO reasons)
- [x] get LTS version when visiting /lts


sivakumar-kailasam:
@MartinMalinda I'm unsure how but would it be possible to research on how the /cuurent paths show up on search engine results first?

MartinMalinda:
But besides that, if the guides and external sources will mostly link to `current`, then it should appear on top.

Maybe we could also have 301 redirect on the latest version to the `current` url.

acorncom:
> I think I read that there is a plan to have no-follow instruction for robots on all but latest project-version

Yes, I think that's what we're going to need to do. We'll likely just handle that via a `robots.txt` file for now ...

sivakumar-kailasam:
Do bots lookup robots file in subpaths too?

acorncom:
@sivakumar-kailasam Don't believe `robots.txt` files are looked up anywhere but root.  We'll need to handle that here as I've currently done: https://github.com/emberjs/website/blob/master/source/robots.txt

acorncom:
@MartinMalinda I imagine you're just starting to recover from EmberConf like the rest of us 😀 But thought I'd poke my head in here to see if we need to do anything else here?

Was talking to Robert Jackson and seems like it might be wiser to shift to a `/release` link instead of a `/current` link here (as current could mean either the latest release or the latest LTS release).  In the long run, we may also want to have a `/lts` link too ... Thoughts?

MartinMalinda:
Yeah, I got excited by Glimmer and invested my "evening/weekend" time into it lately :). I'll try to finish this PR in upcoming days though. Moving from `current` to `release` is fine but I am not sure about `lts`. I guess jsonapi-docs would have to generate isLTS property on project-version model..?

sivakumar-kailasam:
@MartinMalinda I can see how we can add it to the model but we can probably just hardcode it in the app as a config value given that it changes only few times  a year.

app/helpers/url-param.js
```diff
@@ -0,0 +1,12 @@
+import Ember from 'ember';
+
+const { inject } = Ember;
+
+export default Ember.Helper.extend({
+
+  '-routing': inject.service(),
```
sivakumar-kailasam: May be we can use https://github.com/kellyselden/ember-router-service?

config/environment.js
```diff
@@ -2,6 +2,7 @@
 
 module.exports = function(environment) {
   var ENV = {
+    ltsVersion: '2.8.3',
```
sivakumar-kailasam: Should we consider having short circuit to a env variable? That'd probably make it easier to restart app without having to elevate he app.

MartinMalinda: I think I could move straight to https://github.com/emberjs/ember.js/pull/14805. Not sure if it's still under a feature flag though.

MartinMalinda: Nevermind, it does not expose params.

toddjordan:
one thing I noticed is if you go to /release it gets redirected, but if you go to /release/<type>/<name>, it keeps the release in the path.  Is redirecting when just going to /release intentional?

toddjordan:
otherwise looks good.  I'm going to merge and we can update later if need be.

PR #219, opened 2017-05-12
[WIP] Redirect old urls to new

toddjordan:
Will route original urls like `/api/class/Ember.run.html` to `/api/ember/2.12.0/namespaces/Ember.run`


app/routes/class.js
```diff
@@ -0,0 +1,52 @@
+import Ember from 'ember';
+import getLastVersion from '../utils/get-last-version';
+
+export default Ember.Route.extend({
+
+  model(params) {
+    return this.get('store').findRecord('project', 'ember', { includes: 'project-version' })
+      .then((project) => {
+        let versions = project.get('projectVersions').toArray();
+        let lastVersion = getLastVersion(versions);
```
MartinMalinda: In my PR (https://github.com/ember-learn/ember-api-docs/pull/209/files#diff-d1ace59c7fda942c691a37cd9ad5d941R17), lastProjectVersion is a computed property on a project model. I wonder what approach is best. It seems clean as a util but as a computed property it caches properly.

toddjordan: I'll look at using the model computed prop instead of this...

app/utils/hash-to-url.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+  return null;
+
+}
+
+function hashToUrl(window) {
+  if (window && window.location && window.location.hash) {
+    let hashInfo = _parseHash(window.location.hash)
+    if (hashInfo) {
+      return `${window.location.pathname}/${hashInfo.urlType}/${hashInfo.name}?anchor=${hashInfo.name}`
```
toddjordan: possibly will need to append `&show=private,deprecated,protected,inherited` so that old bookmarks show up.  Right now if you have a private (or other non-public) method/event/property bookmarked, it will send you to the page, but won't scroll to the item because its not visible without the url param.

app/routes/class.js
toddjordan: couldn't get it to work without error.  Will look into it more when we bring that commit back.

Issue #222, opened 2017-05-24
Support ember-cli-fastboot rc1


mschinis:
Fastboot rc1 released three days ago made a couple of breaking changes, one of them is using `(instance-)initialisers/browser` structure.

Ember serve fails with:
```
FastBoot build no longer supports /.../ember-learn/ember-api-docs/node_modules/ember-cli-head/app/(instance-)?initializers/browser structure. 
Please refer to www.ember-fastboot.com for a migration path.
```

We should either:
1. do a quick temporary solution to set ember-cli-fastboot dependency to be exact 1.0.0-beta.16 (last known working beta version with this project)
2. Update `ember-cli-head` and `ember-api-docs` to not use these initializers.

Option number 2 is depends on https://github.com/ronco/ember-cli-head/pull/21

PR #223, opened 2017-05-24
Update dependencies

mschinis:
I did a fresh install today and ran across a couple of problems. This fixes all of the issues I had setting up the project.

1. **ember-cli**: Pin it down to minor version bumps. It was upgrading ember-cli to 2.13
2. **ember-load-initializers**: Added ember v2.12 support in v1.0.0
~~3. **ember-cli-fastboot**: Explained [here](https://github.com/ember-learn/ember-api-docs/issues/222)~~


.gitignore
```diff
@@ -6,6 +6,9 @@ terraform.tfstate.backup
 .env
 .envrc
 
+# Editors
```
locks: 👀 could you put this in your global gitignore instead?

package.json
```diff
@@ -50,7 +50,7 @@
     "ember-cli-deploy-gzip": "^0.2.4",
     "ember-cli-document-title": "^0.3.3",
     "ember-cli-eslint": "^3.0.2",
-    "ember-cli-fastboot": "^1.0.0-beta.15",
+    "ember-cli-fastboot": "1.0.0-beta.16",
```
locks: @kratiahuja suggested the latest beta. is it known not to work, or simply not tested?

kratiahuja: Please use `beta.19`. beta builds have no breaking change.

mschinis: 🤔 I remember there was a problem with beta 19, might be an ember-plugin or something. I'll have another look and update the PR.

.gitignore
mschinis: yeap 👍

package.json
kratiahuja: You might have upgraded to the rc builds which has the breaking change. Pin it to beta.19 and it should work.

mschinis:
@kratiahuja updating to beta 19 worked, I think it may have been crashing before because of the loose definition of ember-cli which was grabbing a newer version.

@locks this is good to go

PR #229, opened 2017-06-09
Use async/await instead of andThen() and Promise chains

Turbo87:
/cc @sivakumar-kailasam @toddjordan


app/initializers/set-global-promise.js
```diff
@@ -0,0 +1,7 @@
+import RSVP from 'rsvp';
```
sivakumar-kailasam: Can we add some jsdoc on why the global needs to be set?

Turbo87: done

PR #230, opened 2017-06-09
Use "ember-native-dom-helpers" for acceptance and integration tests

Turbo87:
see https://github.com/cibernox/ember-native-dom-helpers

Disclaimer: This is still not perfect and using jQuery in some cases. Eventually we should use e.g. `ember-test-selectors` to get rid of those jQuery selectors.

/cc @sivakumar-kailasam @mschinis @toddjordan @cibernox

### DIFF
--- a/tests/.eslintrc.js
+++ b/tests/.eslintrc.js
@@ -1,8 +1,6 @@
 module.exports = {
-  env: {
-    'embertest': true
-  },
   globals: {
+    'currentURL': true,
     'selectSearch': true,
     'selectChoose': true
   }
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "ember-load-initializers": "^1.0.0",
     "ember-lodash": "^4.17.1",
     "ember-maybe-import-regenerator": "^0.1.6",
+    "ember-native-dom-helpers": "^0.4.1",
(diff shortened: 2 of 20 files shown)


tests/.eslintrc.js
```diff
@@ -1,8 +1,6 @@
 module.exports = {
-  env: {
```
sivakumar-kailasam: was the removal of this block intentional?

Turbo87: yes, because those helpers (except for `currentURL()`) should not be used anymore when you use `ember-native-dom-helpers`

sivakumar-kailasam: 👍

sivakumar-kailasam:
Will wait for @toddjordan to review this once before we can merge.

tests/.eslintrc.js
cibernox: Actually `ember-native-dom-helpers` also has a currentURL. You can import it and use it.

Turbo87: @cibernox seems to be missing from https://github.com/cibernox/ember-native-dom-helpers#helpers

cibernox: @Turbo87 Fixed.

PR #235, opened 2017-06-16
[WIP] handle all redirects via an express middleware

sivakumar-kailasam:


toddjordan:
@sivakumar-kailasam seems to be working.  Is there more work to do or is it ready to merge?

sivakumar-kailasam:
I haven't moved the redirect from the ember app to the express app yet, so still a WIP. Would be gtg in another couple of days

sivakumar-kailasam:
Resolved as a part of #261

PR #239, opened 2017-06-27
Reintroduce "Handle /release requests"

MartinMalinda:
- [x] make /lts and /release work again 
 - [x] make tests pass
 - [x] make sure it works well with FastBoot
 - [x] add `rel=canonical` meta to pages with versions that match `release` or `lts`
 - [ ] add test coverage for canonicalURL

### DIFF
--- a/app/models/project-version.js
+++ b/app/models/project-version.js
@@ -1,11 +1,8 @@
 import DS from 'ember-data';
 import Ember from 'ember';
-<<<<<<< HEAD
-=======
 import config from 'ember-api-docs/config/environment';
 
 const { computed } = Ember;
->>>>>>> create canonical URL if needed
 
 export default DS.Model.extend({
   version: DS.attr(),
@@ -19,10 +16,10 @@ export default DS.Model.extend({
   'public-namespaces': DS.hasMany('namespace', {async: true}),
   'private-namespaces': DS.hasMany('namespace', {async: true}),
   project: DS.belongsTo('project'),
-  
+
   compactVersion: Ember.computed('version', function() {
     return this.get('version').split('.').slice(0, 2).join('.');
-  })
+  }),
 
   isRelease: computed('project.latestProjectVersion.id', 'version', function() {
     return this.get('project.latestProjectVersion.id') === this.get('version');
--- a/app/models/project-version.js
+++ b/app/models/project-version.js
@@ -1,5 +1,11 @@
 import DS from 'ember-data';
 import Ember from 'ember';
+<<<<<<< HEAD
+=======
+import config from 'ember-api-docs/config/environment';
+
+const { computed } = Ember;
+>>>>>>> create canonical URL if needed
 
 export default DS.Model.extend({
   version: DS.attr(),
@@ -13,7 +19,16 @@ export default DS.Model.extend({
   'public-namespaces': DS.hasMany('namespace', {async: true}),
   'private-namespaces': DS.hasMany('namespace', {async: true}),
   project: DS.belongsTo('project'),
+  
   compactVersion: Ember.computed('version', function() {
     return this.get('version').split('.').slice(0, 2).join('.');
   })
+
+  isRelease: computed('project.latestProjectVersion.id', 'version', function() {
+    return this.get('project.latestProjectVersion.id') === this.get('version');
+  }),
+
+  isLTS: computed(function(){
+    return this.get('version') === config.ltsVersion;
+  }),
 });
--- a/app/models/project-version.js
+++ b/app/models/project-version.js
@@ -25,7 +25,7 @@ export default DS.Model.extend({
     return this.get('project.latestProjectVersion.id') === this.get('version');
   }),
 
-  isLTS: computed(function() {
+  isLTS: computed('version', function() {
     return this.get('version') === config.ltsVersion;
   })
 });
--- a/app/models/project-version.js
+++ b/app/models/project-version.js
@@ -28,7 +28,7 @@ export default DS.Model.extend({
     return this.get('project.latestProjectVersion.id') === this.get('version');
   }),
 
-  isLTS: computed(function(){
+  isLTS: computed(function() {
     return this.get('version') === config.ltsVersion;
-  }),
+  })
 });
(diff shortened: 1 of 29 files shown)


sivakumar-kailasam:
@mschinis will this be ok to have with regards to algolia indexing?

mschinis:
@sivakumar-kailasam yep, I believe we can ignore specific paths as adding both `/release` and `/ember/2.13.3/classes/Ember.Evented/methods/off` would cause duplicate algolia results

MartinMalinda:
I'm for leaving out `/lts` and `/release` from the search results. If someone is doing a search it's likely for some method/event/property which might be version specific.

mschinis:
@MartinMalinda are you going to redirect `/lts` and `/release` to the appropriate url?

MartinMalinda:
@mschinis Probably not so far. So far I am not convinced we should do that. This PR actually goes a little bit into opposite direction.

If you visit `/ember/release` it makes sure that after clicking further links, you stay on `/ember/release`. So that if you then share that link on other places, it keeps linking to the latest version. It also helps SEO if we have one main url that is most dominantly shared. `/release` than should appear on top in search results (hopefully).

I'm open to other suggestions here though. I can be wrong.

sivakumar-kailasam:
@mschinis if we're in a particular version of docs and if we start performing an aloglia search, are results shown to the current version's docs first? 

+1 on not redirecting to specific versions since too many redirects night penalize us on SEO. On a totally unrelated note I tried reaching existing api path structures on api-beta and whatever I tried redirected as expected.

mschinis:
@MartinMalinda 

Agreed on keeping the url the same, but we need to make sure we get this right, as it can actually screw up with SEO. If `/ember/release` and `/ember/2.13.0` offer the same content, then Google will think that the content is duplicate. In the release variant, we would have to set-up `rel="canonical"` tags in the header to point to the original document (2.13.0)

@sivakumar-kailasam yes, the search should only search through the version in the URL, so we might actually need to get algolia to also scrape the current URL. In my previous message above, I thought we were only going to redirect, hence why I mentioned we could ignore specific URLs.

MartinMalinda:
@mschinis there was an idea to have `nofollow` for this but `rel="canonical"` seems like a better choice. I'll see if it can easily be added to this PR otherwise new issue will have to be created.

MartinMalinda:
Ok, it seems like it is possible to populate headData from the metaStore, which contains project versions. So conditional rendering of `<link>` should be possible. I'll finish it tomorrow hopefully.

sivakumar-kailasam:
@MartinMalinda can you rearrange your commits a bit to get a cleaner history?

mschinis:
@MartinMalinda I'm going to add `ember-cli-meta-tags` to add a lot more `<link>` tags to improve our SEO. You can use that plugin to also introduce the canonical link tags.

MartinMalinda:
@mschinis I have the canonical URL mostly done, I'm just improving test coverage. Can refactor it later though.

MartinMalinda:
canonicalURL works now, but there is an issue when `currentURL` is null on the client side. Which causes the canonicalURL to be wiped after booting the app on the client side.

MartinMalinda:
I think this is almost ready to merge. The only thing missing is test coverage for canonicalURL which is hard to do without mirage or any other backend mock. But I'll give it some more thought.

MartinMalinda:
Seems like the only proper way to test the canonicalURL is to use https://github.com/kaliber5/ember-fastboot-app-tests.

toddjordan:
@MartinMalinda do you mind resolving conflicts?  I'd like to get this reviewed/tested.

tests/acceptance/release-lts-project-versions-test.js
```diff
@@ -0,0 +1,67 @@
...
+    return checkLinks('side',
+      '.toc-level-1.modules li:first-child a, .toc-level-1.namespaces li:first-child a, .toc-level-1.classes li:first-child a',
+      versionString);
+  };
+
+  let checkAttributeLinks = versionString => {
+    return checkLinks('attributes', '.attributes a', versionString);
+  };
+
+  let checkIndexListLinks = versionString => {
```
sivakumar-kailasam: not a big deal but this could also be written as, 
```js
let checkIndexListLinks = (versionString) => checkLinks('index-list', '.api-index-filter a', versionString);
```

app/models/project-version.js
```diff
@@ -13,7 +16,16 @@ export default DS.Model.extend({
...
+
   compactVersion: Ember.computed('version', function() {
     return this.get('version').split('.').slice(0, 2).join('.');
+  }),
+
+  isRelease: computed('project.latestProjectVersion.id', 'version', function() {
+    return this.get('project.latestProjectVersion.id') === this.get('version');
+  }),
+
+  isLTS: computed(function() {
```
sivakumar-kailasam: Missed linking to version in the computed definition

MartinMalinda: The version on project-version model does not change, but maybe I should add it anyway for clarity.

tests/acceptance/release-lts-project-versions-test.js
MartinMalinda: 👍  will change that unless the line is too long

app/models/project-version.js
sivakumar-kailasam: Oh yeah that makes sense. Missed the context there :)

tests/acceptance/release-lts-project-versions-test.js
sivakumar-kailasam: Nah this is fine :)

MartinMalinda: Just changed it :D it made me notice this flaw here https://github.com/ember-learn/ember-api-docs/pull/239/commits/f37e79d8d18d014ff5341d326c6b1083060cb7a3#diff-b409b67c33eb7131aee5de1bebe70e5bR14 at least. Passed `selector` variable was actually not used in the helper function.

toddjordan:
we ended up taking a slightly different approach, but currently have release supported.  What we still need to do is add a feature that puts canonical to release in each non-release page. May need some thought on how to do this for pre-2.16 versions that don't match names/modules.

PR #247, opened 2017-07-02
remove changing tagName of link-to and fix inverse relationships

### LINKED ISSUE #213: Links to subroutes are sometimes broken on first load
Links to subroutes like methods, events are sometimes broken.
```
This link-to is in an inactive loading state because at least one of its parameters presently has a null/undefined value, or the provided route name is invalid.
```

I think it would be better to rather not change the tagName of link-to as it is done here:
https://github.com/ember-learn/ember-api-docs/blob/100b9174047ece8290800a60b41fc1a7f86d9e15/app/templates/project-version/class.hbs#L51

Or maybe even switch to ember-href-to to increase performance.

MartinMalinda:
This should fix #213.

The issue there was, that after some time `projectVersion` on the namespace model became undefined.

Maybe because of wrong `inverse` key.

screen shot 2017-07-02 at 13 12 05

I am reloading like crazy and I can't reproduce the issue after this change so  either I am very lucky or it fixed the issue 😀

### DIFF
--- a/app/models/module.js
+++ b/app/models/module.js
@@ -1,12 +1,14 @@
 import ClassModel from './class';
 import DS from 'ember-data';
 
-const { attr } = DS;
+const { attr, belongsTo } = DS;
 
 export default ClassModel.extend({
   submodules: attr(),
   publicclasses: attr(),
   privateclasses: attr(),
   namespaces: attr(),
-  parent: attr()
+  parent: attr(),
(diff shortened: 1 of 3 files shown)


toddjordan:
cool!  want to pull this down and try out before merge, as well as make sure the markup changes didn't affect anything visually

Issue #243, opened 2017-06-29
Links between projects appear to be broken


acorncom:
When clicking the Ember.ArrayProxy link below on https://www.emberjs.com/api-beta/ember-data/2.14.3/classes/DS.RecordArray, my destination link is https://www.emberjs.com/api-beta/ember-data/2.14.3/classes/Ember.ArrayProxy instead of https://www.emberjs.com/api-beta/ember/2.13.3/classes/Ember.ArrayProxy

![screen shot 2017-06-29 at 12 01 40 pm](https://user-images.githubusercontent.com/802505/27702871-c35cbdec-5cc2-11e7-84e9-ba0bf9f51cf7.png)

MartinMalinda:
Clicking https://www.emberjs.com/api-beta/ember-data/2.14.3/classes/DS.RecordArray, I am getting Internal Server Error

MartinMalinda:
The reason for the 500 and FB breaking is.
The issue is here, where `version` ends up undefined. 
https://github.com/ember-learn/ember-api-docs/blob/a23b453bed45d8a04e9e05acc4419446f5af2439/app/adapters/application.js#L33

It looks like a race condition issue. Trying to access the controller befaure it got its model. I think after the merge of #239 it will be easier to get the version from `metaStore` or `routerService`

toddjordan:
looks to be broken until we merge #239  is merged

sivakumar-kailasam:
@toddjordan @MartinMalinda this is working in prod now.

PR #248, opened 2017-07-02
add temporary solution for linking to classes from different project

MartinMalinda:
A temporary and dirty solution for #243.

Ideally, the properties that are currently computed properties on the class model would be provided by `ember-json-api-docs`.

Another option would to transform `uses` and `extends` into relationships. But that could result in extra network requests... or JSON-API could take care of that via `include`?

### DIFF
--- a/app/templates/project-version/classes/class.hbs
+++ b/app/templates/project-version/classes/class.hbs
@@ -8,7 +8,7 @@
     {{#if model.extends}}
       <div class="attribute">
         <span class="attribute-label">Extends:</span>
-        <span class="attribute-value">{{#link-to 'project-version.classes.class' model.projectVersion.version model.extends}}{{model.extends}}{{/link-to}}</span>
+        <span class="attribute-value">{{#link-to 'project-version.classes.class' model.extendedClassProjectName model.extendedClassVersion model.extends}}{{model.extends}}{{/link-to}}</span>
       </div>
     {{/if}}
     {{#if model.uses}}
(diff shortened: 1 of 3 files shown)


toddjordan:
Thanks @MartinMalinda will pull it down and try it out.  Can you fix up the eslint errors for CI?

MartinMalinda:
@toddjordan Fixed!

app/templates/project-version/classes/class.hbs
```diff
@@ -8,7 +8,7 @@
     {{#if model.extends}}
       <div class="attribute">
         <span class="attribute-label">Extends:</span>
-        <span class="attribute-value">{{#link-to 'project-version.classes.class' model.projectVersion.version model.extends}}{{model.extends}}{{/link-to}}</span>
+        <span class="attribute-value">{{#link-to 'project-version.classes.class' model.extendedClassProjectName model.extendedClassVersion model.extends}}{{model.extends}}{{/link-to}}</span>
```
toddjordan: the name is available from the route.  we should be able to pass it to the template instead of guessing in the model object.  you can get it from `transition.params['project-version'].project` in the model hook.

toddjordan: wasn't talking about version, but name

MartinMalinda: Ah, well then it would not solve anything, we'd be linking to from `ember-data` to `ember-data` same as before, having broken link. We need to be able to link from `ember-data` to `ember` back and forth.

toddjordan: ok, I get what's happening now.  this is the link to a class that the current one is exteded from cross project.  disregard 👍

PR #249, opened 2017-07-04
[wip]Support fastboot plus hash params

toddjordan:
Since hash params aren't sent to the server, and since route model hooks aren't run client side with fastboot.  I'm "cheating" here and inserting a script that converts the hash for legacy urls.  I'm doing this via implementing `contentFor` in an in-repo-addon.

### DIFF
--- a/lib/hash-to-query/index.js
+++ b/lib/hash-to-query/index.js
@@ -1,0 +1,41 @@
+/* eslint-env node */
+'use strict';
+
+module.exports = {
+  name: 'hash-to-query',
+
+  isDevelopingAddon: function() {
+    return true;
+  },
+
+  contentFor (type, config) {
+    if (type === 'body') {
+      return `
+<script type='text/javascript'>
+  let hash = window.location.hash;
+  if (hash && hash.length > 1) {
+    let pathName = window.location.pathname;
+    console.log(pathName);
+    if (!pathName.includes('/methods/') &&
+        !pathName.includes('/properties/') &&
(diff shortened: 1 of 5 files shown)


sivakumar-kailasam:
@toddjordan may be we can rewrite the url in fastboot beforeMiddleware? Because the urls within docs will never have old hashed urls, only old third party references which will hit fastboot first time.

toddjordan:
@sivakumar-kailasam hashes wouldn't even get there.  They aren't sent to the server.

sivakumar-kailasam:
Ah forgot about that :/

lib/hash-to-query/index.js
```diff
@@ -0,0 +1,41 @@
...
+
+  isDevelopingAddon: function() {
+    return true;
+  },
+
+  contentFor (type, config) {
+    if (type === 'body') {
+      return `
+<script type='text/javascript'>
+  let hash = window.location.hash;
```
MartinMalinda: by using `let` we are probably breaking IE10, but I don't know if that concerns us. The usage of IE 10 among Ember devs is probably near 0%.

MartinMalinda: or the contentFor in this case gets transpiled by Babel?

toddjordan: good point will check on ie10 + ember 2.14

toddjordan:
Need to fix tests for this, and want to try out in a heroku preview since I've seen some fastboot differents between local and deployed.

Issue #246, opened 2017-07-01
Improve performance


MartinMalinda:
I did a little bit of performance checkup and profiling today. I believe some things could significantly improve performance.

- `link-to`'s are very expensive and there's a lot of them. Moving to `ember-href-to` would have a very positive impact on perf. Href-to has some issues with dynamic changes of query-params but I think it's worth a try.
- `href-to` could be then memoized for further perf increase
- background reloading in Ember data could be turned off, since data are mostly static anyway
- `tether.js` is a bit expensive
- install https://github.com/salsify/ember-cli-dependency-lint
 - remove lodash?

toddjordan:
performance seems to be pretty good atm.  Maybe do some payload analysis and profiling in the future but for now we will continue as is.  Thanks for looking at it.

Issue #220, opened 2017-05-22
Refactor api-docs route structure


mschinis:
### The problem

Our current solution for indexing the api-docs in Algolia relies on Algolia's docsearch. We have [this JSON](https://github.com/algolia/docsearch-configs/blob/master/configs/emberjs_versions.json) that describes how the docsearch scraper reads the website and follows the various links to construct a full index of all the api docs.

In order to write an equivalent json configuration, we need to use classes that the scraper will identify as level1/level2/etc of the hierarchy. An example hierarchy would be the following:
`API (lvl0) > Modules (lvl1) > ember (lvl2) > Ember.ContainerDebugAdapter (lvl3)`
for this url `https://www.emberjs.com/api-beta/ember/2.12.0/modules/ember-debug`

Right now, we have a `current` class on each element in the hierarchy, that helps us:
1. Style and give better visual feedback to the person reading the docs
2. Extract the full hierarchy using the docsearch scraper

*Example of what it looks like on current docs*

![](http://take.ms/e7pam)

Similarly to the above, navigating to this url `https://www.emberjs.com/api-beta/ember/2.12.0/modules/ember-debug` should have applied a class that the scraper can identify as being lvl1. Investigating this further, this is because the lvl1 links are essentially plain `<a>` tags with no logic to apply an identifiable class.

*Example of beta api-docs*
![](http://take.ms/skbyj)

### The solution

The proposed solution here would be to use ember's `{{link-to}}` and delegate all of that logic to it.
Since ember does the url detection very well, and handles any class updates depending on the url on its own, I think it's very appropriate to use hand all of this to ember.

Given the url structure in the `router.js` looks like the following
```javascript
    this.route('namespace', {path: '/namespaces/:module'}, itemRoutes);
```

A link-to lvl1 anchor that automatically updates the current class is not possible.

I propose we restructure the routes, to something like the following:

```javascript
this.route('namespaces', function () {
    this.route('namespace', { path: '/:module' });
});
```

Making the above restructure would allow us to use the `link-to` to create something like the following:
```hbs
{{#link-to 'namespaces' }}
  Modules
{{/link-to}}
```

Ofcourse, this will require us to also update any existing links to the `namespace` route to point to `namespaces.namespace`.

Similarly to the above example of `/namespace/:module`, we will need to do the same for;
1. `/modules/:module`
2. `/classes/:class`

[...]

PR #252, opened 2017-07-04
Replace link-to component with href-to helper

MartinMalinda:
This replaces `link-to` with `href-to`. 
It provides roughly 40% decrease in rendering time on `api-beta/ember/2.13.3/namespaces/Ember`.

Not that noticeable on initial load, but it is noticeable on further transitions

relates to #246

### DIFF
--- a/app/components/table-of-contents.js
+++ b/app/components/table-of-contents.js
@@ -1,6 +1,15 @@
 import Ember from 'ember';
 
+const { inject, computed } = Ember;
+
 export default Ember.Component.extend({
+
+  transitionService: inject.service('transition'),
+  currentParam: computed.alias('transitionService.newThirdParam'),
+  currentNamespaceID: computed.alias('currentParam.namespace'),
+  currentModuleID: computed.alias('currentParam.module'),
+  currentClassID: computed.alias('currentParam.class'),
+
   actions: {
     toggle(type) {
       this.$('ol.toc-level-1.' + type).slideToggle(200);
--- a/app/mixins/scroll-tracker.js
+++ b/app/mixins/scroll-tracker.js
@@ -4,15 +4,16 @@ const { inject, Mixin } = Ember;
 
 export default Mixin.create({
 
-  scrollPositionReset: inject.service(),
+  transitionService: inject.service('transition'),
 
   actions: {
     willTransition(transition) {
-      this.get('scrollPositionReset').scheduleReset(transition);
+      this.get('transitionService').trackDynamicParams(transition);
+      this.get('transitionService').scheduleScrollReset(transition);
     },
 
     didTransition() {
-      this.get('scrollPositionReset').doReset();
+      this.get('transitionService').doReset();
     }
   }
 });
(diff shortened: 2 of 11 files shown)


toddjordan:
@MartinMalinda can you look at clearing conflicts?

MartinMalinda:
@toddjordan I'm resolving #239 atm. I will resolve conflicts here after that's merged - it will introduce more conflicts probably.

locks:
@MartinMalinda could we use the router service instead?

MartinMalinda:
@locks I am not sure what you mean, this PR uses the router service

locks:
@MartinMalinda I meant making our own `href-to` that uses the Router service, but it's fine to land this and iterate.
Are you able to fix the conflicts?

MartinMalinda:
@locks I see. I am looking into the router service more now and I most likely get rid of some logic and use it more.

We could use the service to generate URLs and to apply `active` class on links but without `href-to` there would be no global handling of clicking the links, so we'd have to probably have `onclick` on every link.

I'll resolve conflicts here and refactor a bit once #239 is merged so I don't have to resolve twice.

MartinMalinda:
I think `ember-href-to` will probably use the router service soon anyway.

mschinis:
@MartinMalinda Be careful with this approach as algolia is using the `active` class on each link to identify the current link. Check out my PR description here:

https://github.com/ember-learn/ember-api-docs/issues/220

MartinMalinda:
@mschinis I've added logic to add the `active` class on links but I will definitely double checks things are working correctly and maybe add some test coverage if there isn't already any.

Using the `isActive` function from the router service will possibly further make it error-proof

acorncom:
One question here re replacing `link-to`. I appreciate the idea of increased speed (Id love to see things go faster). How does that play in with our idea of the API docs being a canonical reference app that represents Ember best practices? Seems a bit strange to hold it up as a good example but then also be swapping out link-to ... :grinning: Thoughts?

MartinMalinda:
I think by providing the official `Router` service, which @locks recommends above, the things are changing a little bit.

The service provides a function for URL generation, provides an `isActive` class, the only missing piece is handling clicks on such links. `Ember-href-to` uses a global event listener for that, but a regular `action` could be used.

```
{{#with (get-url 'foo.bar' someModel) as |url|}}
<a href={{url}} class={{if (is-active url) "active"}} {{action (transition url)}}>My link</a>
{{/with}}
```

It's more verbose but also more explicit.

MartinMalinda:
I could also adjust this PR a bit, using `https://github.com/rwjblue/ember-router-helpers`. It now provides a `route-params` helper. Usage is like so

```{{#with (route-params 'parent.child') as |routeParams|}}
      <a href="{{routeParams.url}}" class="{{if routeParams.isActive 'active' 'inactive'}}">Blah</a>
    {{/with}}

toddjordan:
Rendering is much improved in newer Embers, and our app performance isn't bad, so putting this one on ice until rendering performance becomes more of a pressing need.  Thanks for the work and may pull this back up eventually.

Issue #251, opened 2017-07-04
support legacy ember data urls


toddjordan:
example:  https://emberjs.com/api/data/classes/DS.Adapter.html

toddjordan:
Taking a similar approach to other legacy urls (create a custom route and transition to appropriate new route)

Currently blocked by #250 though

toddjordan:
Fixed by #253

PR #253, opened 2017-07-05
Add ember data legacy url routing

toddjordan:
addresses #251


app/routes/data-module.js
```diff
@@ -0,0 +1,43 @@
+import Ember from 'ember';
```
toddjordan: quite a bit a duplicated code around these.  Will refactor and try to combine code + efficiency.

PR #254, opened 2017-07-06
rely on ember-cli-head to set document title

### LINKED ISSUE #244: <title> is always the same in FastBoot
It seems like in FastBoot the `<title>` is always "Ember API Documentation" and the correct title is set on client side only

MartinMalinda:
fixes #244

At this point, both `ember-cli-head` and `ember-cli-document-title` is used to set the `title`. Ideally `ember-cli-document-title` would be removed but its function for collecting tokens is quite useful so I'm not removing it so far.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -4,7 +4,6 @@
 
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <title>Ember API Documentation</title>
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link rel="shortcut icon" href="{{rootURL}}assets/images/favicon.png">
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -5,7 +5,7 @@ const { set, inject } = Ember;
 
 export default Ember.Route.extend({
   headData: inject.service(),
-  title: function(tokens) {
+  title(tokens) {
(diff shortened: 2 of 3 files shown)


MartinMalinda:
I'll be looking at FastBoot tests later today so while I am at it, I may add a test that `title` is correctly set in FastBoot

sivakumar-kailasam:
@MartinMalinda would replacing document title with [ember-page-title](http://tim-evans.github.io/ember-page-title/) help? It does play well with ember-cli-head & fastboot.

MartinMalinda:
@sivakumar-kailasam Probably. This setup works as well even though it is not that clean. `document.title` is set twice in the browser. On the other hand, I like having `title` in the `head.hbs` so that all dynamic head data are in one place.

`ember-page-title` would allow more flexible title formats, but I am not sure we need that at this point?

sivakumar-kailasam:
Yeah had similar thoughts but thought i'd run it by you :) Can you rebase your code with master? It has changes that'd make the heroku deployment pass.

MartinMalinda:
@sivakumar-kailasam done, but now Travis is failing for some reason 🤔

sivakumar-kailasam:
Travis just needed a restart. I've seen it happen when chrome download sha checks fail in travis randomly. Tested via a review app on heroku, so I'll merge it.

Issue #186, opened 2017-03-23
Error template doesn't show up in some scenarios


sivakumar-kailasam:
For both cases mentioned below we'll have to start by handling failures in the fetch that happens in the application adapter.

- In case there's a typo in the request, http://ead-sk.herokuapp.com/api-new/ember/2.12.0/namespaces/Embera, then fastboot gives us this old school error. 
<img width="960" alt="screenshot 2017-03-24 01 05 19" src="https://cloud.githubusercontent.com/assets/604117/24266749/f96d7b88-102d-11e7-9c3b-cea2c1e1563e.png">

- When user is offline & he switches to a url that isn't cached yet, the failed requests aren't handled.  It would be fancy to show a template that gives out this info about being offline & not being covered by service workers on this particular url. The [navigator api](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine) is probably useful here.
<img width="958" alt="screenshot 2017-03-24 01 06 22" src="https://cloud.githubusercontent.com/assets/604117/24266828/3170622a-102e-11e7-9e8e-d636b7d5ba04.png">

toddjordan:
We should try to fix using route error substates, if fastboot supports.

toddjordan:
Also, we need to support older patch releases, because people bookmark them.  @sivakumar-kailasam can we access previous patch data?

sivakumar-kailasam:
@toddjordan we never showed actual patch release's docs on the UI via the version select. So I stopped generating docs for anything except the latest patch release. May be we can redirect to latest patch? If not I can generate older patch releases as well.

acorncom:
I'm seeing this same issue when visiting http://emberjs.com/api-beta/ember/2.12.0/namespaces/Ember  Looks like it's related to not having `2.12.0` in the index

http://emberjs.com/api-beta/ember/2.12.2/namespaces/Ember works fine

toddjordan:
@acorncom known issue we talked about in the meeting:  https://github.com/ember-learn/ember-api-docs/issues/179

toddjordan:
I think https://github.com/ember-fastboot/fastboot/issues/134 is relevant to this issue

toddjordan:
Tried the suggestion in  ember-fastboot/fastboot#134 and didn't affect the behavior

toddjordan:
Here's the actual stacktrace:
```
Error: TransitionAborted
[...]

PR #262, opened 2017-07-09
doing error transition in the redirect hook when class isn't found

toddjordan:
per https://github.com/ember-fastboot/ember-cli-fastboot/issues/202

see #186

### DIFF
--- a/app/routes/project-version/classes/class.js
+++ b/app/routes/project-version/classes/class.js
@@ -25,20 +25,27 @@ export default Ember.Route.extend(ScrollTracker, {
   find(typeName, param) {
     return this.store.find(typeName, param).catch(() => {
       return this.store.find('namespace', param).catch(() => {
-        return this.transitionTo('project-version'); // class doesn't exist in new version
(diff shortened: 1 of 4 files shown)


app/routes/project-version/classes/class.js
```diff
@@ -25,20 +25,27 @@ export default Ember.Route.extend(ScrollTracker, {
   find(typeName, param) {
     return this.store.find(typeName, param).catch(() => {
       return this.store.find('namespace', param).catch(() => {
-        return this.transitionTo('project-version'); // class doesn't exist in new version
+        return Ember.RSVP.resolve({ isError: true });
```
toddjordan: So fastboot gives back a 500 when you transition during the model hook or afterModel hook, so I need to communicate to the redirect hook that I want to send this to our 404 page.

app/templates/404.hbs
```diff
@@ -0,0 +1,10 @@
+<article class="whoops">
```
toddjordan: Using the same template as error.hbs (without the logic).  Open to some design love.

PR #263, opened 2017-07-10
Compact version in url

### LINKED ISSUE #179: Visiting older version of docs while on a newly introduced class causes errors that aren't handled
Visit `ember/2.12.0/classes/Ember.Engine` and switch version on the left to any 1.x version. The app immediately try to makes a request for a class that doesn't exist so the app is left in a broken state.

sivakumar-kailasam:
Fixes #179


config/target.js
```diff
@@ -1,9 +1,19 @@
...
+  'last 1 Chrome versions',
+  'last 1 Firefox versions',
+  'last 1 Safari versions'
+];
+
+if (process.env.EMBER_ENV === 'development' || process.env.EMBER_ENV === 'test') {
+  browsers = [
     'last 1 Chrome versions',
-    'last 1 Firefox versions',
-    'last 1 Safari versions'
```
toddjordan: no safari?

toddjordan:
awesome!

config/target.js
sivakumar-kailasam: I assumed all of us favor chrome during development 🙈

PR #269, opened 2017-07-11
Fix header CSS

Turbo87:
/cc @sivakumar-kailasam


sivakumar-kailasam:
still seems to have issues on some screens. Please use the review app to test these changes

PR #279, opened 2017-07-12
Use SVG icons instead of fontello icon font

Turbo87:
/cc @locks @sivakumar-kailasam


sivakumar-kailasam:
@Turbo87 is it possible to use svg-jar in https://github.com/ember-learn/glimmer-styleguide/blob/master/src/ui/components/main-footer/template.hbs ? that's used on the rest of the website

Turbo87:
probably? I don't know how the rest of the website works

sivakumar-kailasam:
@locks already made it the footer elsewhere. But my question was whether we can use the addon on a glimmer repo.

Issue #287, opened 2017-07-14
Main container chapter has variable width


makepanic:
When comparing https://emberjs.com/api/ember/2.14/namespaces/Ember with https://emberjs.com/api/ember/2.14/classes/Ember.Controller one can see that the chapter content width is different.
This is most likely caused by `main.container .content .chapter` only using `max-width` (800px) instead of using a fixed width.

mikoscz:
@sivakumar-kailasam I think that we can close this one 😄

PR #295, opened 2017-07-20
Fix problem with chapter container width

mikoscz:
Fix problem with width chapter container described in: https://github.com/ember-learn/ember-api-docs/issues/287


locks:
Thank you @mikoscz!

PR #296, opened 2017-07-20
Added more space when showing our spinner

### LINKED ISSUE #284: Make loading template taller so transition isn't as abrupt
At the moment the loading template is very short, putting the footer almost in the middle of the screen.
When the page finally loads, it suddenly expands, being a bit jarring.

acorncom:
Fixes #284 

**Before**
before

**After**
after

### DIFF
--- a/app/styles/components/_loading-spinner.scss
+++ b/app/styles/components/_loading-spinner.scss
@@ -1,6 +1,6 @@
 .loading-spinner {
   text-align: center;
-  margin: 5em auto;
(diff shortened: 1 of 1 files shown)


toddjordan:
lgtm, though for some reason I can't load the heroku preview...

sivakumar-kailasam:
Review apps get their config from staging & staging has a fastly url configured in env which causes that error :/. We can remove that var once we start using production instance on website.

Issue #276, opened 2017-07-11
Duplicate method/properties/events when showing 'inherited'


ascot21:
When you check the 'inherited' (or 'protected') checkbox when viewing the API docs for a module, you see duplicated methods, properties and events.  See screenshot below:

![screen shot 2017-07-11 at 10 34 10 am](https://user-images.githubusercontent.com/578560/28081645-c8c04ff4-6624-11e7-9ddd-63e2920f41c6.png)

toddjordan:
Likely showing documentation up the inheritance change.  We should show the local documentation if both exist.

jenweber:
@toddjordan is file path depth a reasonable proxy for which inheritance is more local? Is there a better option besides filtering things based on that?

toddjordan:
@jenweber this will probably need some work in https://github.com/ember-learn/ember-jsonapi-docs to provide enough information for the client to determine which item to use (or more likely to filter it out).  We need to look at what's available there.  Most likely we have enough info (since we did in the old ruby-based api site I think)

jenweber:
Agreed. I spent some time last night inspecting the network requests to see what I could use, and came to the conclusion that I needed something external, whether an inheritance tree or different query results/structure.

I'll check it out.

jenweber:
The last commit with the api docs in the website repo was `dfb2591970183e15b7d16dc12ebf1573c885917d`

Digging around there and attempting to read Ruby, it seems like methods/properties had a flag added of overwritten_from or extended_from that informed display filtering, and it was based on a consequential ordering of the YUIdoc info.

In cases of duplicates, I'm going to try preferring the 2nd item and see if that does the trick.

api_docs.rb
```ruby
    def classitems
      unless @classitems
        @classitems = {}

        # Go from root up
        parents = ([extends] + uses.reverse).compact
        parents.each do |parent|
          parent.classitems.each do |type, items|
            items.each do |_, item|
              item = item.clone
              @classitems[item['itemtype']] ||= {}
              existing = @classitems[item['itemtype']][item['name']]
              if existing
                next if existing['class'] == item['class']
                item['overwritten_from'] = existing
              else
                item['extended_from'] = parent
              end
              @classitems[item['itemtype']][item['name']] = item
            end
          end
        end

        self.class.data.classitems.each do |item|
          next if item['class'] != self.name

          item = item.clone
          @classitems[item['itemtype']] ||= {}
          if existing = @classitems[item['itemtype']][item['name']]
[...]

PR #301, opened 2017-07-23
bugfix: duplicate properties (#276)

jenweber:
Addresses https://github.com/ember-learn/ember-api-docs/issues/276

Whenever there are duplicate methods, the first one in the filtered list is the "most local," as far as I can see. If I'm reading the code right, it appears that this is how it was handled in the past as well.

One piece is missing: What would a good test be? It would be easy to write a test that asserts that there aren't duplicates, but I'm most concerned about the source data ordering changing somehow and _that_ not being caught, resulting in the "less specific" documentation being shown. Suggestions? I can add a test before a merge. I'm PR'ing it now for easy testing/comparison/advice by a reviewer.

cc @sivakumar-kailasam @toddjordan


sivakumar-kailasam:
@jenweber Looking at the response data from server for a url where we see duplicates https://emberjs.com/api/ember-data/2.14/classes/DS.JSONAPIAdapter I noticed a few things we can use from the response for getting to know the most local one,

```
```

sivakumar-kailasam:
@jenweber Looking at a shorter version of the [response data from server](https://ember-api-docs.global.ssl.fastly.net/json-docs/ember-data/2.14.6/classes/ember-data-2.14.6-DS.RESTAdapter-428fda3f07.json) for [json api adapter](https://emberjs.com/api/ember-data/2.14/classes/DS.JSONAPIAdapter), one case where we see duplicates, it looks like we have an extra piece of information we could use to figure out the most local entity with confidence.


<details>
<summary>Short JSON API doc </summary>

```json
{
  "data": {
    "id": "ember-data-2.14.6-DS.RESTAdapter",
    "type": "class",
    "attributes": {
      "extends": "DS.Adapter",
      "file": "addon/adapters/rest.js",
      "uses": [
        "DS.BuildURLMixin"
      ],
      "methods": [
        {
          "name": "createRecord",
          "file": "addon/adapters/rest.js"
        },
        {
          "name": "createRecord",
          "file": "addon/adapter.js"
        }
      ]
    },
    "relationships": {
      "parent-class": {
        "data": {
          "id": "ember-data-2.14.6-DS.Adapter",
          "type": "class"
        }
      },
      "descendants": {
        "data": [
          {
            "type": "class",
            "id": "ember-data-2.14.6-DS.JSONAPIAdapter"
          }
        ]
      },
      "module": {
        "data": {
          "id": "ember-data-2.14.6-ember-data",
          "type": "module"
        }
      },
      "project-version": {
        "data": {
          "id": "ember-data-2.14.6",
          "type": "project-version"
        }
      }
    }
  }
}
```

</details>


The module has a `file` attribute that is also present in the event/methods/properties model. May be grouping by name and picking the one with the same filename as module, if not the filename of the base class of the module & so on(unsure if this extended use case exists) might be more accurate. 

On the matter of tests, may be an acceptance test would suffice. We could probably use the rest adapter or something similar to verify unique items & their sources (may be `data-test-file` to indicate their source?)

Sorry I couldn't put up this info sooner, hadn't investigated this issue until now 🙁

jenweber:
Don't merge yet. This needs another careful code review and a test.

sivakumar-kailasam:
@jenweber I merged from master so that we could see your changes on the review app. Added a WIP prefix to keep us from hitting that green button :)

jenweber:
@sivakumar-kailasam ok good, you were making me nervous, haha

sivakumar-kailasam:
Let's ship it

PR #306, opened 2017-07-26
[RFC] Adds classes to target with algolia

mschinis:


toddjordan:
@mschinis does this let the algolia selectors work?

mschinis:
@toddjordan I'm adding css selectors and tweaking their script to make it work. 
This should be the only css selector missing. The others I had to change them slightly.

PR #308, opened 2017-07-28
Modules refactor

acorncom:


acorncom:
Applies the https://github.com/ember-cli/ember-modules-codemod  Do we want to apply this or wait until 2.16 lands? /cc @locks 

I'm inclined to use it as we can use this app to help teach folks about how to make the change ...

sivakumar-kailasam:
@acorncom there's a eslint config to ensure that direct ember imports would cause eslint failures. We should probably add it along with this PR. @Turbo87 might be aware of it.

Turbo87:
> there's a eslint config to ensure that direct ember imports would cause eslint failures

the rule only warns about usage of the old shims from `ember-cli-shims` but it does not warn about `import Ember from 'ember'` yet because there are still legitimate use cases for that (e.g. `Ember.testing`).

> We should probably add it along with this PR

seems like a somewhat unrelated change and should be done in a separate PR to not block this one

acorncom:
@Turbo87 I pushed a few buttons, it's really easy to do 😀 Thanks for all your work on that ...

locks:
Oops, I also just pushed a couple of buttons. Nice work!

Issue #271, opened 2017-07-11
Get urls to specific methods/props/events by clicking anchor image on heading


toddjordan:
(missing feature from legacy ruby app)

toddjordan:
Links to individual items follow the format ex:
https://emberjs.com/api/ember/2.14/classes/Ember.Route/methods/activate?anchor=activate

Blocked by issue #273

toddjordan:
cc @alexander-alvarez

mikoscz:
I'm taking this one ;)

alexander-alvarez:
:beers:

alexander-alvarez:
@mikoscz have you started on this? I can tackle today or tomorrow if not

PR #311, opened 2017-07-30
Implementing anchor image

alexander-alvarez:
https://github.com/ember-learn/ember-api-docs/issues/271

Currently looks like so on hover:
image

Clicking on it routes you to the `?anchor=<method>`

Also made use of the `ember-svg-jar` for `fa-pencil` since it was using `partial` still & I looked at that to see how SVGs where being used, only to realize it was the ugly :duck: of all the `SVGs in the app.


toddjordan:
Thanks for the PR, Looks good!  Two minor comments.  The svg looks to be offset a bit low.  I think it would look better raised a couple of pixels.   

Also, in the old app I think the image itself was also linkable.  it isn't a big deal ( can ship without), but if its easy to do we should.

alexander-alvarez:
@toddjordan Not a problem. I changed the styles after looking at github's implementation. Let me know if want any adjustments. (do you think the iconography is too large?)

PR #312, opened 2017-07-30
[WIP] Split out module name from access (DOM)

mschinis:
This splits out the module name header from the access span. 
It will help getting rid of the "public", "private" from the algolia search results.

This should be the final change needed before we can get the search back up.

I'll re-run the docsearch scraper once again once this is on staging, but we should be good to go


mschinis:
Needs a bit of tweaking for the methods section:

http://take.ms/3W4ob

mschinis:
@toddjordan  Can we merge this in?

Issue #285, opened 2017-07-14
Rename Modules to Packages


locks:
We currently have a header in the table of contents that reads "Modules", as well as a route segment "/modules". This should be renamed to "Packages"  and "/packages".

I'm wondering if we can just rename it as it's very early days, or we have to include aliases?

amyrlam:
👋 explored the repo some for the first time today! i can take a stab at this one

sivakumar-kailasam:
As long as `/modules` can redirect to `/packages` so that we don't break bookmarked/indexed urls out there.

PR #315, opened 2017-08-02
[WIP] Rename /modules to /packages (#285)

amyrlam:
Resolves: #285

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -45,7 +45,7 @@ AppRouter.map(function() {
     // project-version.module => project-version.modules.module
     //    routes/project-version/module   =>  routes/project-version/modules/module
     //    routes/project-version/module/* =>  routes/project-version/modules/module/*
-    this.route('modules', function() {
+    this.route('modules', {path: '/packages'}, function() {
       this.route('module', {path: '/:module'}, itemRoutes);
     });
     // this.route('module', {path: '/modules/:module'}, itemRoutes);
@@ -64,9 +64,9 @@ AppRouter.map(function() {
     }
   });
   this.route('class', {path: '/classes/:class'});
-  this.route('module', {path: '/modules/:module'});
+  this.route('module', {path: '/packages/:module_id'});
   this.route('data-class', {path: '/data/classes/:class'});
-  this.route('data-module', {path: '/data/modules/:module'});
+  this.route('data-module', {path: '/data/packages/:module'});
 });
 
 export default AppRouter;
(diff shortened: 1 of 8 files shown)


amyrlam:
very WIP:

- not sure about failing tests

- want to have /modules/* manual input in url bar redirect to /packages/*, not sure how given that it's not a `beforeModel`

- getting this strange error, ideas?

```alam@tesla ember-api-docs (amy/module-package-taketwo) $ ember serve
Warning: failed to stat /Users/alam/src/ember-api-docs/tmp/broccoli_merge_trees-input_base_path-7W5XmxWe.tmp/0/ember-api-docs/controllers/project-version/packages/package/events.js
```

jenweber:
@amyrlam 
I see the "warning: failed to stat..." error all the time. I think it's safe to ignore for the scope of this.

I resolved a test failure for "lists all subpackages on the package page" by changing line 28:
`let numberSubModules = Object.keys(container.get('submodules')).length;` Submodules should be plural.

Here's one possible way to keep routes like "modules" around. As time goes on, we'll accumulate more of these urls that we need to support. You could create a wildcard in the router like `this.route('catch-all', { path: '*:' });`. Then in the catch-all, handle exceptions like modules/packages:
```
import Ember from 'ember';

export default Ember.Route.extend({
  beforeModel(transition) {
    let intended = transition.intent.url
    if (intended.match(/modules/i)) {
      let repaired = intended.replace('modules', 'packages')
      this.replaceWith(repaired)
    } else {
      this.transitionTo('404')
    }
  }
});
```

@sivakumar-kailasam @toddjordan what do you think about having a catch all route in the api docs?

jenweber:
Cc @locks regarding wildcard routes as a way to handle legacy urls. What do you think?

amyrlam:
@jenweber heya, thanks for the feedback. please keep it coming! 

i'm just getting back from a ~2wk vacation with almost no interweb access. will revisit this PR friday/this weekend

jenweber:
@amyrlam learning team feedback on my idea above is, wildcard routes can be messy and we shouldn't use them for this case.

In order to support both `/packages` and `/models` routes, the suggestion from the learning team chat is that the app should have 2 routes, one for packages and one for models. The only job of /modules will be to redirect. So the router would look something like 

```
    this.route('packages', {path: '/packages'}, function() {
      this.route('module', {path: '/:module'}, itemRoutes);
    });
    this.route('modules')
```

The `/:module` dynamic segment can stay the same, since it's internal.

Does that make sense? I wasn't quite sure what you meant when you mentioned some issues with beforeModel in your comment above, and I haven't really dug into the routing structure of the app much myself. I think you can possibly use the `redirect` hook to catch the full intended transition.

toddjordan:
Thanks so much for this work, but we took a simpler approach by just renaming the labels and not changing our internal references.  That way if we change our minds back it won't be a big deal :-P

Issue #317, opened 2017-08-03
code block layout bug in mobile, can't copy & paste


jenweber:
Line numbers somehow push the code block below the column on small screens.

Also you can't copy and paste the code blocks with touch controls (at least on iphone 5s). I added some CSS in a previous PR that prevents the line numbers column from being copied, but that shouldn't bleed over. I think copy & paste is related to the cursor type of pointer vs text.

![screen shot 2017-08-03 at 2 51 23 pm](https://user-images.githubusercontent.com/16627268/28938416-fbe86634-785b-11e7-8153-710f67f7ab57.png)

jenweber:
I'll take this one. Almost fixed.

Issue #274, opened 2017-07-11
Header/Footer Styling on Mobile


toddjordan:

toddjordan:
cc @Turbo87

toddjordan:
partially addressed by https://github.com/ember-learn/ember-api-docs/pull/269

jenweber:
PR https://github.com/ember-learn/ember-api-docs/pull/320

toddjordan:
@jenweber do you think this is closable?

jenweber:
Yes. Closed.

PR #320, opened 2017-08-05
Mobile friendlier header, footer, and code blocks

jenweber:
Closes https://github.com/ember-learn/ember-api-docs/issues/317 and maybe https://github.com/ember-learn/ember-api-docs/issues/274

**Sidebar bug:** I brought in the media query breaks just a bit, to align the sidebar behavior with the container. This means it jumps into mobile arrangement sooner. 

**Responsive header/footer:** The main problem was that all the responsive CSS ported over from the main site wasn't being applied because that class was was not included in the html.

**Coderay block separating from line numbers:** just a global-ish css rule on tables that didn't make sense.

**Coderay block selection**: in the more global CSS was a `z-index: -1` applied to table layouts, which had an unexpected effect on code blocks. Setting the z index back to normal for coderay tables fixes the cursor problem. TBD whether code can be copied & pasted in mobile.

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -12,14 +12,10 @@ body {
   width: 100%;
 }
 
-.header {
-  z-index: 2;
-}
-@include media($medium-large-screen-up) {
+@include media($large-screen-up) {
   .header {
     top: 0;
     width: 100%;
-    position: fixed;
   }
 
   main.container {
@@ -29,7 +25,6 @@ body {
   main.container .content {
     position: relative;
     margin-right: 0;
-    margin-top: $top-spacing;
     margin-bottom: 0;
     overflow:auto;
     padding: $base-spacing;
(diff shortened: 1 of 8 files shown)


sivakumar-kailasam:
@jenweber let me know if it was because of page layout in which case this looks good to merge.

jenweber:
I'll change the breakpoints to match. Good catch.

jenweber:
@sivakumar-kailasam I have updated the header/footer breakpoints to match the rest of the website. It still needs another pair of eyes just to be sure. 

I have also replaced any stray breakpoints to be the variables in `_grid-settings.scss`. I changed the header nav to be relatively positioned. It was fixed and none of the rest of the site works that way. If that was intentional, we just need to ignore the final commit.

If it all looks right for you too, I think this is good to go.

sivakumar-kailasam:
Looks great now @jenweber. The header being fixed must've been a bug.

Issue #318, opened 2017-08-05
Autoscroll to methods section is broken


sivakumar-kailasam:
Github links to the methods are broken
https://emberjs.com/api/ember/2.14/namespaces/Ember.computed/methods

Reported via https://github.com/emberjs/website/issues/2971

sivakumar-kailasam:
//cc @MartinMalinda

jaredgalanis:
It looks like across api-docs there are no anchor values for the method, properties, or events tabs. I can take this on.

sivakumar-kailasam:
@jaredgalanis there's an open PR #321 for this one, so may be one of the other open issues?

jaredgalanis:
Ah, got it. Did not see catch that. Yep, I'll take a look.

EndangeredMassa:
Should this be closed? It appears to have been resolved by: https://github.com/ember-learn/ember-api-docs/pull/321

sivakumar-kailasam:
Looks like we broke this again 😿

toddjordan:
Tried listed scenario and works currently.

PR #321, opened 2017-08-10
[Bugfix]Fix autoscroll to methods issue

akashdsouza:
If this looks ok, I can add the same to /methods under classes and modules(perhaps in a mixin?).

### DIFF
--- a/app/mixins/scroll-tracker.js
+++ b/app/mixins/scroll-tracker.js
@@ -1,5 +1,7 @@
 import Mixin from '@ember/object/mixin';
 import { inject as service } from '@ember/service';
+import $ from 'jquery';
+import config from 'ember-api-docs/config/environment';
 
 export default Mixin.create({
 
@@ -11,7 +13,18 @@ export default Mixin.create({
     },
 
     didTransition() {
-      this.get('scrollPositionReset').doReset();
+      this._super();
+      let isAnchorEmpty = window && window.location && window.location.search === '?anchor='
+      if ((typeof FastBoot === 'undefined') && isAnchorEmpty ) {
+        let elem = $('#methods');
+        let offset = (elem && elem.offset && elem.offset()) ? elem.offset().top : null;
+        if (offset) {
+          const navMenuHeight = $('header').outerHeight();
+          $(config.APP.scrollContainerSelector).scrollTop(offset - navMenuHeight - 10);
+        }
+      } else {
+        this.get('scrollPositionReset').doReset();
+      }
     }
   }
 });
--- a/app/mixins/scroll-tracker.js
+++ b/app/mixins/scroll-tracker.js
@@ -14,17 +14,15 @@ export default Mixin.create({
 
     didTransition() {
       this._super();
-      let isAnchorEmpty = window && window.location && window.location.search === '?anchor='
-      if ((typeof FastBoot === 'undefined') && isAnchorEmpty ) {
+      if ((typeof FastBoot === 'undefined') && window.location.search === '?anchor=' ) {
         let elem = $('#methods');
-        let offset = (elem && elem.offset && elem.offset()) ? elem.offset().top : null;
+        let offset = elem.offset() ? elem.offset().top : 0;
         if (offset) {
-          const navMenuHeight = $('header').outerHeight();
-          $(config.APP.scrollContainerSelector).scrollTop(offset - navMenuHeight - 10);
+          $(config.APP.scrollContainerSelector).scrollTop(offset - 10);
+          return;
         }
-      } else {
-        this.get('scrollPositionReset').doReset();
       }
+      this.get('scrollPositionReset').doReset();
     }
   }
 });
--- a/app/components/ember-anchor.js
+++ b/app/components/ember-anchor.js
@@ -10,10 +10,9 @@ export default AnchorComponent.extend({
     let qp = this.get('anchorQueryParam');
     let qpVal = this.get(get(this, 'attrs.a') ? 'a' : `controller.${qp}`);
     let elem = $(`[data-${qp}="${qpVal}"]`);
-    let offset = (elem && elem.offset && elem.offset()) ? elem.offset().top : null;
+    let offset = elem.offset() ? elem.offset().top : 0;
     if (offset) {
-      const navMenuHeight = $('header').outerHeight();
-      $(config.APP.scrollContainerSelector).scrollTop(offset - navMenuHeight - 10);
+      $(config.APP.scrollContainerSelector).scrollTop(offset);
     }
   }
 });
--- a/app/routes/project-version/namespaces/namespace.js
+++ b/app/routes/project-version/namespaces/namespace.js
@@ -1,6 +1,7 @@
 import ClassRoute from '../classes/class';
+import ScrollTracker from 'ember-api-docs/mixins/scroll-tracker';
 
-export default ClassRoute.extend({
+export default ClassRoute.extend(ScrollTracker, {
   templateName: 'project-version/classes/class',
 
   model(params, transition) {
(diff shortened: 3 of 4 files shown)


sivakumar-kailasam:
Fixes #318

app/routes/project-version/namespaces/namespace/methods.js
```diff
@@ -1,5 +1,21 @@
...
+import $ from 'jquery';
+import config from 'ember-api-docs/config/environment';
 
 export default Route.extend({
-  templateName: 'methods'
+  templateName: 'methods',
+  actions: {
+    didTransition() {
+      this._super();
+      if ((typeof FastBoot === 'undefined') && isEmpty(this.get('controller.anchor'))) {
```
MartinMalinda: I think I've mentioned it before somewhere. `isEmpty(this.get('controller.anchor'))` this will probably result in scroll jump unless there is anchor specifically set up. Which is probably not a good behavior.

@sivakumar-kailasam @locks 
Do we need to have backward compatibility on this? Or could we come up with a more specific anchor for scrolling to methods?

sivakumar-kailasam: Unfortunately we do. There's just so many old links out there outside our website that refer to the old doc urls. We can probably do an analysis of google analytics metrics to see the trends few months down the line and make changes but until then we have to keep at this.

MartinMalinda: @akashdsouza 
Oh okay. Then I think we need to specifically check for `null` or `""` (or both?) so it only detects just `?anchor=`

MartinMalinda: Hopefully, that will work and there won't be a need to look into `location`.

akashdsouza: Alright. I will update the check.
Should I use a data-attribute instead of id for scrolling to the tab?

MartinMalinda: I think ID is fine.

akashdsouza: Updated

akashdsouza:
This might have to wait until #322  is merged

app/mixins/scroll-page.js
```diff
@@ -0,0 +1,23 @@
...
+      this._super();
+      if ((typeof FastBoot === 'undefined') && this.get('controller.anchor') === '' ) {
+        let elem = $('#methods');
+        let offset = (elem && elem.offset && elem.offset()) ? elem.offset().top : null;
+        if (offset) {
+          const navMenuHeight = $('header').outerHeight();
+          $(config.APP.scrollContainerSelector).scrollTop(offset - navMenuHeight - 10);
+        }
+      } else {
+        this.get('scrollPositionReset').doReset();
```
MartinMalinda: Why is the reset needed if there's no anchor?

app/routes/project-version/classes/class/methods.js
```diff
@@ -1,5 +1,6 @@
 import Route from '@ember/routing/route';
```
MartinMalinda: I think the correct place to put the Mixin would be `project-version/classes/class.js` and `project-version/modules/module.js` and `project-version/namespaces/namespace.js`. That should cover all cases.

MartinMalinda:
Works well, but the Mixin should be put in different places as I've suggested above, otherwise, the scroll jump may not happen in various places.

I think we should also consider updating these query params
https://github.com/ember-learn/ember-api-docs/blob/7020a59b84a63a71c88a5da969f4950d726650a1/app/templates/project-version/classes/class.hbs#L50

There is a query-param reset like this `(query-params anchor="")`. I think the reason for that was so there would not be a random jump to some method - but now it cases a scroll jump to the tab element. `(query-params anchor=undefined)` should solve that.

app/mixins/scroll-page.js
akashdsouza: To retain the behavior available here: 
> https://github.com/ember-learn/ember-api-docs/blob/7943dba1683b30b89249ae11118cae93aad79f1b/app/mixins/scroll-tracker.js#L14

Earlier, as there was no didTransition in `/methods`, didTransition above would have been called.

app/routes/project-version/classes/class/methods.js
akashdsouza: There is already a scroll-tracker mixin in these routes to reset scrolling. How would this need to be handled? Remove the scroll reset altogether as it is present in scroll-page mixin?

MartinMalinda: I think we could actually add this logic to the Scroll Tracker mixin itself. It is already at the right places - maybe it will just need to be added to `namespaces/namespace.js`.

app/mixins/scroll-tracker.js
```diff
@@ -11,7 +13,18 @@ export default Mixin.create({
     },
 
     didTransition() {
-      this.get('scrollPositionReset').doReset();
+      this._super();
+      let isAnchorEmpty = window && window.location && window.location.search === '?anchor='
+      if ((typeof FastBoot === 'undefined') && isAnchorEmpty ) {
```
akashdsouza: Does this seem hacky?

app/mixins/scroll-tracker.js
```diff
@@ -11,7 +13,18 @@ export default Mixin.create({
...
+        let elem = $('#methods');
+        let offset = (elem && elem.offset && elem.offset()) ? elem.offset().top : null;
+        if (offset) {
+          const navMenuHeight = $('header').outerHeight();
+          $(config.APP.scrollContainerSelector).scrollTop(offset - navMenuHeight - 10);
```
akashdsouza: Since header is no longer a fixed element on the page, do we still need to take navMenuHeight into consideration?

app/mixins/scroll-tracker.js
MartinMalinda: Getting anchor from the controller cannot be done from these places?
If we are not in FastBoot, it is safe to reach into `window.location.search` right away. `(typeof FastBoot === 'undefined') && window.location.search === '?anchor=')` should be fine.
This will also break if we have more than 1 queryParam, but this is probably fine. So far there no plans for adding more.

app/mixins/scroll-tracker.js
MartinMalinda: > Having a little space above the method doesn't look all bad

There will be less space if we remove it :). But let's do it, I think the anchor is gonna be more clear.

app/mixins/scroll-tracker.js
akashdsouza: Yes. anchor is only available in child route's(`methods`) controller.
I will update the check

akashdsouza:
Updated

app/mixins/scroll-tracker.js
```diff
@@ -11,6 +13,15 @@ export default Mixin.create({
     },
 
     didTransition() {
+      this._super();
+      if ((typeof FastBoot === 'undefined') && window.location.search === '?anchor=' ) {
+        let elem = $('#methods');
+        let offset = (elem && elem.offset && elem.offset()) ? elem.offset().top : null;
```
MartinMalinda: This is also overly defensive. If it's JQuery object, it is safe to assume that `elem.offset` will be a function even if the element is not found.

`let offset = elem.offset() ? olem.offset().top : 0` should be fine.

MartinMalinda:
Works well 💯 The offset check could be simplified but it's also fine as is.

app/mixins/scroll-tracker.js
akashdsouza: Updated. Sorry about the delay

toddjordan:
looks good, thanks @akashdsouza !!!

PR #323, opened 2017-08-17
styles: parameter and return type tweaks

knownasilya:
Make it more obvious where the type ends and the description begins. I didn't check this locally.

Looks something like this:

screen shot 2017-08-17 at 10 30 16 am

### DIFF
--- a/app/styles/components/_class-field-desc.scss
+++ b/app/styles/components/_class-field-desc.scss
@@ -17,11 +17,14 @@
   font-size: $base-font-size;
 }
 
-.return-type {
+.class-field-description--link .return-type,
+.return .return-type,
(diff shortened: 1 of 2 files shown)


toddjordan:
excellent!

sivakumar-kailasam:
@toddjordan should we change the return type text color to something brighter?

knownasilya:
I was going to do a dark orange, but the links are almost the same color. If links had underlines, we could do that.

sivakumar-kailasam:
@knownasilya  may be increase the font weight to begin with?

knownasilya:
Of the type or the link?

sivakumar-kailasam:
The return type.

PR #326, opened 2017-08-19
WIP | Do not show private API checkboxes

### LINKED ISSUE #303: Hide quick access to private APIs
On the left-hand side on TOC we need to show deprecated classes always without the need to enable the checkbox. The checkbox for `Show Private/Deprecated` needs to removed. We need to add a new query param that will let us show private classes/namespaces/modules which will be false/hidden by default (may be `?show-private-toc=true`).

In addition we need to remove the private checkbox on the methods/properties index on right side. If someone needs to view private info, they can use the existing query param (`?show=private`) to do view it.

//@Kapeli - fyi

locks:
Fixes #303.

- [x] Hide checkboxes
- [ ] Show deprecated classes
  - [ ] Mark classes as deprecated visually
- [ ] Add query param for hidden private class toggle

### DIFF
--- a/app/templates/components/table-of-contents.hbs
+++ b/app/templates/components/table-of-contents.hbs
@@ -30,8 +30,3 @@
     </ol>
   </li>
 </ol>
-<label class="toc-private-toggle">
-  {{input type="checkbox" checked=showPrivateClasses class='private-deprecated-toggle'}}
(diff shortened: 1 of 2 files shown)


toddjordan:
Putting this on ice for now, since its been a few years.

PR #328, opened 2017-08-23
Test for meta-store service and minor updates

lifeart:
* compactVersion safe get
* tests for services/meta-store.js


tests/unit/services/meta-store-test.js
```diff
@@ -2,10 +2,86 @@ import { moduleFor, test } from 'ember-qunit';
...
+  let service = this.subject();
+  service.initializeStore(availableProjectVersions, projectRevMap);
+  assert.ok(service.get('availableProjectVersions.ember.length'));
+});
+
+test('it should return correct fullVersion from compact', function(assert) {
+  let service = this.subject();
+  service.initializeStore(availableProjectVersions, projectRevMap);
+  assert.equal(service.getFullVersion('ember', '2.14'), '2.14.1');
+  assert.equal(service.getFullVersion('ember', '2.1'), '2.14.1');
```
toddjordan: seems like a bad assert here.

lifeart: @toddjordan that should I update?

toddjordan:
I've merged these changes manually, along with some additional minor fixes to work with the current codebase.  Thanks for the contribution and sorry its taken so long.
image

PR #329, opened 2017-08-23
Priority for already loaded record

lifeart:
if we already have loaded recored - do not reload it from server


app/services/filter-data.js
```diff
@@ -1,7 +1,7 @@
 import Service from '@ember/service';
 
 export default Service.extend({
-  showInherited: false,
+  showInherited: true,
```
sivakumar-kailasam: we wouldn't want to change this behaviour for now

lifeart: @sivakumar-kailasam if I set it to `false`, 4 tests fail and this is magic for me..

lifeart: @sivakumar-kailasam done

lifeart:
@Turbo87

app/routes/project-version/classes/class.js
```diff
@@ -64,7 +69,7 @@ export default Route.extend(ScrollTracker, {
 
   serialize(model) {
     return {
-      class: model.get('name')
+      class: get(model, 'name')
```
Turbo87: Why did you change this?

lifeart: @Turbo87 `class` route alteady has `import { set, get } from '@ember/object';`, why not to use them?

Turbo87: because they are much harder to read IMHO and only make sense if `model` was a POJO and not an `Ember.Object`

lifeart: Rolled-back

toddjordan:
I might be off base here, but doesn't ember-data implicitly do caching?  findRecord should load existing items in the store if already loaded.

lifeart:
@toddjordan looks like it's depends on adapter configuration - https://github.com/emberjs/data/blob/master/addon/-legacy-private/system/store.js#L520

toddjordan:
Verified, additional calls to that json are loaded from disk cache, reducing the need for this update.

PR #336, opened 2017-09-01
feat: sticky method, property, and event headers

knownasilya:
sticky

I'm 95% satisfied with this, the other 5% is if the section is really small I'd prefer not to show a sticky, but the addon just doesn't support it yet. I submitted a feature request here https://github.com/kaliber5/ember-sticky-element/issues/4


sivakumar-kailasam:
@knownasilya Will this be sticky on mobile as well?

knownasilya:
Yes

simonihmig:
Fyi: the feature request has been implemented in https://github.com/kaliber5/ember-sticky-element/pull/12 (though differently, as was discussed). @knownasilya please let me know in case it does not work for you as expected!

toddjordan:
Thanks for this fix, I tried this out on a few browser types and seems to work, but I'd like to use the current version of this addon (0.2.1) as it removes jquery in newer versions, plus your fix request will be implemented.  @knownasilya do you have any bandwidth to update?

knownasilya:
Sure, I'll have a look today.

knownasilya:
@toddjordan should be all set.

Issue #335, opened 2017-08-31
Anchors/Links within Class description don't scroll


alexander-alvarez:
Because of https://github.com/ember-learn/ember-api-docs/blob/master/app/templates/project-version/classes/class.hbs#L41

Eg. when you click `pushPayload` from https://www.emberjs.com/api/ember-data/2.14/classes/DS.Store
you get this url in your address https://www.emberjs.com/api/ember-data/2.14/classes/DS.Store#method_pushPayload but no scroll occurs.
(if you refresh, it does scroll though)

![image](https://user-images.githubusercontent.com/11779993/29935325-58ab6bae-8e4c-11e7-8416-da422fadf066.png)

Could we hook into the behavior that parses the refresh logic into the proper `?anchor=<>` url and https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onhashchange or similar concept?

locks:
Seems fixed!

PR #337, opened 2017-09-01
[FIX] Handle hash change in url

akashdsouza:
https://github.com/ember-learn/ember-api-docs/issues/335

cc @MartinMalinda

### DIFF
--- a/app/instance-initializers/hash-to-query-redirect.js
+++ b/app/instance-initializers/hash-to-query-redirect.js
@@ -1,0 +1,18 @@
+import hashToQuery from 'hash-to-query';
+
+export function initialize(application) {
+  if (typeof FastBoot === 'undefined') {
+    window.onhashchange = function() {
+      let router = application.lookup('service:router');
+      let redirectToUrl = hashToQuery(window.location.hash, window.location.pathname);
+      if (redirectToUrl) {
+        router.transitionTo(redirectToUrl);
+      }
+    }
+  }
+}
+
+export default {
+  name: 'hash-to-query-redirect',
+  initialize
+};
--- a/lib/hash-to-query/index.js
+++ b/lib/hash-to-query/index.js
@@ -8,33 +8,8 @@ module.exports = {
     return true;
   },
 
-  contentFor (type, config) {
-    if (type === 'body') {
-      return `
-<script type='text/javascript'>
-  var hash = window.location.hash;
-  if (hash && hash.length > 1) {
-    var pathName = window.location.pathname;
-    if (pathName.indexOf('/methods/') === -1 &&
-        pathName.indexOf('/properties/') === -1 &&
-        pathName.indexOf('/events/') === -1) {
-      var type = hash.slice(1, hash.indexOf('_'));
-      var name = hash.slice(hash.indexOf('_') + 1, hash.length);
-      var anchor = '?anchor=' + name + '&show=inherited,protected,private,deprecated';
-      var newPath = pathName;
-      if (type === 'method') {
-        newPath = pathName + '/methods/' + name;
-      } else if (type === 'property') {
-        newPath = pathName + '/properties/' + name;
-      } else if (type === 'event') {
-        newPath = pathName + '/events/' + name;
-      }
-      window.location.href = window.location.origin + newPath + anchor;
-    }
-  }
-</script>
-`
-    }
-    return '';
+  included(app) {
+    this._super.included.apply(this, arguments);
+    this.import('vendor/hash-to-query.js');
   }
 };
--- a/lib/hash-to-query/vendor/hash-to-query.js
+++ b/lib/hash-to-query/vendor/hash-to-query.js
@@ -1,0 +1,35 @@
+(() => {
+  /* globals define */
+  function hashToQuery(hash, pathName) {
+    if (hash && hash.length > 1) {
+      if (pathName.indexOf('/methods/') === -1 &&
+          pathName.indexOf('/properties/') === -1 &&
+          pathName.indexOf('/events/') === -1) {
+        let type = hash.slice(1, hash.indexOf('_'));
+        let name = hash.slice(hash.indexOf('_') + 1, hash.length);
+        let anchor = '?anchor=' + name + '&show=inherited,protected,private,deprecated';
+        let newPath = pathName;
+        if (type === 'method') {
+          newPath = pathName + '/methods/' + name;
+        } else if (type === 'property') {
+          newPath = pathName + '/properties/' + name;
+        } else if (type === 'event') {
+          newPath = pathName + '/events/' + name;
+        }
+        return newPath + anchor;
+      }
+    }
+  }
+  if (typeof FastBoot === 'undefined') {
+    let redirectToUrl = hashToQuery(window.location.hash, window.location.pathname);
+    if (redirectToUrl) {
+      window.location.href = window.location.origin + redirectToUrl;
+    }
+  }
+  define('hash-to-query', function () {
+    'use strict';
+    return {
+      default: hashToQuery
+    }
+  });
+})()
--- a/tests/unit/initializers/hash-to-query-redirect-test.js
+++ b/tests/unit/initializers/hash-to-query-redirect-test.js
@@ -1,0 +1,33 @@
+import Ember from 'ember';
+import { initialize } from 'ember-api-docs/instance-initializers/hash-to-query-redirect';
+import { module, test } from 'qunit';
+import destroyApp from '../../helpers/destroy-app';
+import hashToQuery from 'hash-to-query';
+
+module('Unit | Instance Initializer | hash to query redirect', {
+  beforeEach() {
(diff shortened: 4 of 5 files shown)


akashdsouza:
I'm also trying to write a couple of acceptance tests for both behavior. However, I'm not sure as how to defer the asserts

MartinMalinda:
As long as it will trigger page refresh it's not gonna be testable.

Some ideas:

 - Make `hashToQuery` return the new url, instead of doing the redirect directly
 - Create a `shim` for this function
 - Add an initializer, import this function shim there, add an eventListener for `hashchange`
 - Lookup the router service and in the `hashchange` callback, do `router.transitionTo(newURL)` (it should accept URL afaik)

akashdsouza:
I have made some changes. @MartinMalinda do these changes look ok?

lib/hash-to-query/vendor/hash-to-query/index.js
```diff
@@ -0,0 +1,32 @@
+(() => {
```
MartinMalinda: `lib/hash-to-query/vendor/hash-to-query/index.js` that looks a bit strange file path, is that intentional?

app/instance-initializers/hash-to-query-redirect.js
```diff
@@ -0,0 +1,20 @@
+import hashToQuery from 'hash-to-query';
+
+export function initialize(application) {
+  let redirectToUrl = hashToQuery();
+  if (redirectToUrl) {
+    window.location.href = window.location.origin + redirectToUrl;
```
MartinMalinda: I would do this first check already in the vendor file, it will happen sooner.

tests/unit/initializers/hash-to-query-redirect-test.js
```diff
@@ -0,0 +1,25 @@
...
+    });
+  },
+  afterEach() {
+    Ember.run(this.appInstance, 'destroy');
+    destroyApp(this.application);
+  }
+});
+
+// Replace this with your real tests.
+test('it works', function(assert) {
```
MartinMalinda: It would be good to have some input/output tests for the `hashToQuery` function here. (unless the test coverage for it is elsewhere already)

lib/hash-to-query/vendor/hash-to-query/index.js
```diff
@@ -0,0 +1,32 @@
...
+  /* globals define */
+  function hashToQuery() {
+    var hash = window && window.location && window.location.hash;
```
MartinMalinda: This is overly defensive. `hashToQuery` should be called when `window.location` exists.

app/instance-initializers/hash-to-query-redirect.js
MartinMalinda: It will also need `typeof FastBoot === undefined` check there.

tests/unit/initializers/hash-to-query-redirect-test.js
MartinMalinda: Simple acceptance test would be cool as well.

lib/hash-to-query/vendor/hash-to-query/index.js
akashdsouza: I just looked up ember-router-service-polyfill addon for reference. But then again, this is just an in-repo addon. Any suggestions?

MartinMalinda: I am only wondering about placement of the file `lib -> hash-to-query -> vendor -> hash-to-query -> index.js`.

app/instance-initializers/hash-to-query-redirect.js
akashdsouza: > I would do this first check already in the vendor file, it will happen sooner.

I don't understand. Wouldn't we need to have this check before setting location href?

akashdsouza: > It will also need typeof FastBoot === undefined check there

If this check can be moved to vendor file, can't `window` check also be placed along with that

MartinMalinda: > If this check can be moved to vendor file, can't window check also be placed along with that

If `FastBoot` is undefined, `window` will be defined. No need to double check for `window`.

tests/unit/initializers/hash-to-query-redirect-test.js
akashdsouza: Added tests for hashToQuery

tests/unit/initializers/hash-to-query-redirect-test.js
```diff
@@ -0,0 +1,33 @@
...
+
+  // you would normally confirm the results of the initializer here
+  assert.ok(true);
+});
+
+test('hashToQuery method works', function(assert) {
+  initialize(this.appInstance);
+
+  let redirectToUrl = hashToQuery('method', '/class');
+  assert.equal(redirectToUrl, '/class?anchor=method&show=inherited,protected,private,deprecated', 'hashToQuery function works');
```
MartinMalinda: try these tests with some real data scenarios:

```
let redirectToUrl = hashToQuery('#method', '/class');
  assert.equal(redirectToUrl, '/class?anchor=method&show=inherited,protected,private,deprecated', 'hashToQuery function works');
  assert.equal(hashToQuery('#method_pushPayload', '/ember-data/2.14/classes/DS.Store'), '/ember-data/2.14/classes/DS.Store/methods/pushPayload?anchor=pushPayload&show=inherited,protected,private,deprecated');
  assert.equal(hashToQuery('#method_pushPayload', '/ember-data/2.14/classes/DS.Store/methods/createRecord'), '/ember-data/2.14/classes/DS.Store/methods/pushPayload?anchor=pushPayload&show=inherited,protected,private,deprecated');
  assert.equal(hashToQuery('#method_push', '/ember-data/2.14/classes/DS.Store/methods/createRecord'), '/ember-data/2.14/classes/DS.Store/methods/push?anchor=push&show=inherited,protected,private,deprecated');
```

Only the second one passes. The others fail because they contain `/methods/`. Either the `hashToQuery` function needs to be updated or a new function needs be made that shares some of the logic.

MartinMalinda:
I'm looking into this a little bit more the situation on the initial load of the app and the internal `hashchange` are a bit different.

Maybe reusing the `hashToQuery` function is not the best way and just having two different function would be better.

I hacked this together really quick and it works:

```js
import hashToQuery from 'hash-to-query';
import Inflector from 'ember-inflector';

const {inflector} = Inflector;

export function initialize(application) {
  if (typeof FastBoot === 'undefined') {
    window.onhashchange = function() {
      const { hash } = window.location;
      if (hash.includes('#') && hash.includes('_')) {
        const [type, anchor] = hash.replace('#', '').split('_');
        const router = application.lookup('service:router');
        const currentRoute = router.get('currentRouteName');
        const baseRoute = currentRoute.split('.').splice(0, 3).join('.');
        const subRoute = `${inflector.pluralize(type)}.${type}`;
        router.transitionTo(`${baseRoute}.${subRoute}`, {
          queryParams: { anchor }
        });        
      }
    }
  }
}

export default {
  name: 'hash-to-query-redirect',
  initialize
};
```

MartinMalinda:
We can also update to 2.15 and the polyfill does not need to be installed

MartinMalinda:
@akashdsouza would it be fine for you if I pull your commit and create a new PR with my changes?

akashdsouza:
@MartinMalinda Sure. Go ahead

lib/hash-to-query/index.js
```diff
@@ -8,33 +8,8 @@ module.exports = {
     return true;
   },
 
-  contentFor (type, config) {
```
toddjordan: is this really going to work for direct url access, such as say I have a bookmarked link for something with `#method_toString`?

toddjordan: if we are taking out the need for contentFor, then this doesn't need to be an in-repo addon.  just make it a util function.

lib/hash-to-query/vendor/hash-to-query.js
```diff
@@ -0,0 +1,35 @@
...
+        } else if (type === 'property') {
+          newPath = pathName + '/properties/' + name;
+        } else if (type === 'event') {
+          newPath = pathName + '/events/' + name;
+        }
+        return newPath + anchor;
+      }
+    }
+  }
+  if (typeof FastBoot === 'undefined') {
```
toddjordan: what's this for?  Should never enter this if I don't think...

tests/unit/initializers/hash-to-query-redirect-test.js
toddjordan: We need to verify that legacy links with hashes work (as the currently do today)

toddjordan:
tried to deploy a heroku preview app for this but got an error running yarn: 
image

toddjordan:
Going to drop support for legacy method/property links, making this fix moot.

Issue #307, opened 2017-07-26
Legacy redirect may lead to strange result


MartinMalinda:
This URL: https://www.emberjs.com/api/classes/String.html
Leads to:
https://www.emberjs.com/api/ember/2.14.1/classes/String?show=inherited%2Cprotected

Which is probably not a correct route and the correct methods there are only available under the `private` filter.

jenweber:
I also noticed some weird redirect from http://emberjs.com/api/classes/Ember.Router.html#method_map

Which is linked as `map()` in the [guides](https://guides.emberjs.com/v2.15.0/routing/defining-your-routes/). It sends users to https://emberjs.com/api/ember/2.15.0/classes/Ember.Router/methods/map?anchor=map&show=inherited,protected,private,deprecated

Related to https://github.com/ember-learn/ember-api-docs/issues/303

jenweber:
PR https://github.com/ember-learn/ember-api-docs/pull/339

toddjordan:
we route most of these to the landing page now, and are dropping support for routing lagacy properties and methods.

PR #339, opened 2017-09-09
WIP bugfix: hide private API when following guides/legacy links (#307)

jenweber:
**WIP don't merge.**

Addresses https://github.com/ember-learn/ember-api-docs/issues/307

Based on conversation about hiding private API by default (https://github.com/ember-learn/ember-api-docs/issues/303), I think it makes sense that legacy urls should not show private api by default either.

Currently, links from the Guides redirect to showing all QPs (inherited, protected, private, deprecated), so new learners would see private API a lot. I fixed this by changing the legacy url processing.

### DIFF
--- a/lib/hash-to-query/index.js
+++ b/lib/hash-to-query/index.js
@@ -20,7 +20,12 @@ module.exports = {
         pathName.indexOf('/events/') === -1) {
       var type = hash.slice(1, hash.indexOf('_'));
       var name = hash.slice(hash.indexOf('_') + 1, hash.length);
-      var anchor = '?anchor=' + name + '&show=inherited,protected,private,deprecated';
+      var anchor = '?anchor=' + name;
+      var queryParams = window.location.search;
+      var matchedQP = queryParams.match(/inherited|protected|private|deprecated/gi);
+      if (matchedQP) {
+        anchor += '&show=' + matchedQP.join(',');
+      }
       var newPath = pathName;
       if (type === 'method') {
         newPath = pathName + '/methods/' + name;


toddjordan:
This would kill people's bookmarks

jenweber:
@toddjordan because people have bookmarks that include the QPs? I'll read this a little more closely and see if I can account for those.

toddjordan:
Yes people have bookmarks to private things.  I guess it's a question of whether to support it

jenweber:
Broken links are the worst. I didn't really think it through all the way, just saw that legacy links get all the boxes checked and that the Guides are linking people to private APIs. I'll give it another shot.

toddjordan:
Yeah. Agreed it's confusing linking from guides. We should def. convert thing guide links to the new URL structure

jenweber:
@toddjordan I think maybe this fixes it?

As of my latest commit:
- if you have an old link that goes to something private, deprecated, or protected, you will see those QP in your redirect (and only those, so if you have a "private" QP and no others, you will only see only private API)
- if you have an old link with no QP specified, you will see only inherited by default (even if no inherited methods are available or you were linking to private API)

Examples:
**Legacy link with hash & QP:** http://localhost:4200/classes/Ember.Router.html?show=private#method_map
**Redirects to:** https://emberjs.com/api/ember/2.15.0/classes/Ember.Router/methods/map?anchor=map&show=private
**Shows:** only private API is checked.

**Legacy hashed link without QP:** http://localhost:4200/classes/Ember.Router.html?e#method_map
Redirects to: http://localhost:4200/ember/2.15.0/classes/Ember.Router/methods/map?anchor=map
**Shows:** Only inherited is checked via fallback to application wide defaults.

**Legacy links without a hash or QP:** http://localhost:4200/classes/String.html
**Redirects to** http://localhost:4200/ember/2.15.0/classes/String
**Shows:** Only inherited is checked via fallback to application wide defaults. Since String has only private methods, none are shown by default.

**Legacy links with QP and without a hash:** http://localhost:4200/classes/String.html?show=private
**Redirects to** http://localhost:4200/ember/2.15.0/classes/String?show=private
**Shows:** Only private is checked.

**Legacy private API links with private QP:**  http://localhost:4200/classes/Ember.Router.html?show=private#method__deserializeQueryParam
**Redirects to** http://localhost:4200/ember/2.15.0/classes/Ember.Router/methods/_deserializeQueryParam?anchor=_deserializeQueryParam&show=private
**Shows:** Only private is checked.

**Legacy private API links missing QP:**  http://localhost:4200/classes/Ember.Router.html#method__deserializeQueryParam
**Redirects to** http://localhost:4200/ember/2.15.0/classes/Ember.Router/methods/_deserializeQueryParam?anchor=_deserializeQueryParam
**Shows:** Only inherited is checked via fallback to application wide defaults. Does not show specified private method!

jenweber:
@toddjordan I also experimentally tried to fix https://github.com/ember-learn/ember-api-docs/issues/340 so let me know if you want me to remove that change & squash the commits.

jenweber:
I just realized I am making the assumption that the old API site used QPs and they were named the same. Is that true?

jenweber:
I'm just going to close this and do something totally different after the links have settled. It's partially irrelevant from replacing all the links in 2.16 guides

Issue #333, opened 2017-08-30
Using an API version number is not always loading the right version


RustyToms:
I'm so happy that we can filter the docs by version number now, I don't know when that was added but it is really great thanks!

However, sometimes the wrong version of the docs are loading. For example, when I go to https://emberjs.com/api/ember/2.2/namespaces/Ember I generally see that the 2.12 docs load, even though the url is unchanged. I even got 1.12 to load once. Similarly, https://emberjs.com/api/ember/2.1/namespaces/Ember loads the 2.14 docs for me.

I am using Chrome Version 61.0.3163.59 (Official Build) beta (64-bit) on a Mac

akashdsouza:
This is happening because of this line
> https://github.com/ember-learn/ember-api-docs/blob/master/app/services/meta-store.js#L38
Versions 2.12.2 also matches string 2.2 causing it to be the first element of the array which leads to the wrong version being loaded

akashdsouza:
Includes of `includes`, may be we can use `startsWith` ?

MartinMalinda:
I think earlier we had a data structure of versions. Something like this:
```js
versions: [{
  compactVersion: "2.1"
   fullVersion: "2.1.1"
}, {
  compactVersion: "2.14"
  fullVersion: "2.14.3"
}]
```
I think I used that in my PR back then: https://github.com/ember-learn/ember-api-docs/pull/239/files#diff-d1ace59c7fda942c691a37cd9ad5d941R38. This is more explicit and reliable approach. Unfortunately that PR is old and unmergeable right now.

jenweber:
PR https://github.com/ember-learn/ember-api-docs/pull/348

PR #348, opened 2017-09-28
fix loose matching of versions (#333)

jenweber:
In the live site, choosing some `1.x` from the version dropdown will sometimes bring you to the wrong version docs or leave the dropdown in a weird state. In the old code, 1.2 would match 2.11.2, for example. This PR does a more careful match.

Issue: https://github.com/ember-learn/ember-api-docs/issues/333

### DIFF
--- a/app/services/meta-store.js
+++ b/app/services/meta-store.js
@@ -35,7 +35,13 @@ export default Service.extend({
 
   getFullVersion(projectName, compactProjVersion) {
     const availProjVersions = this.get(`availableProjectVersions.${projectName}`);
-    return availProjVersions.filter(v => v.includes(compactProjVersion))[0];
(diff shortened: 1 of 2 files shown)


jenweber:
Yeah, I can add one. It was a pretty bad UI bug... making life even harder for the 1.0 crew

jenweber:
@acorncom I added some tests! Let me know if they should be different somehow. I don't use unit testing much.

jenweber:
Let's hold off a week before merging this in. Too many moving parts with 2.16 happening

sivakumar-kailasam:
@jenweber looks good to merge, can we merge now?

jenweber:
@sivakumar-kailasam I think so. Just delayed because I had no idea what would need to change for 2.16 but the version numbering is all the same

Issue #350, opened 2017-10-05
QC changes to the stack for RFC 176


jenweber:
- [x] make sure that we don't break the api docs for older versions of Ember.
- [x] make version switching work
- [ ] fix so many tests

I tested it for 2.16, 2.11, 1.8, and 1.0. Layout is fine. Styling is meh.

The dropdown menu nav breaks. Fixed in https://github.com/ember-learn/ember-api-docs/pull/351

If you navigate to a 2.16 module, then try to go to 1.8, you get this:

```Error: Assertion Failed: The URL '/ember/1.8/modules/@ember/component' did not match any routes in your application```

Going from 2.11 class to 2.18 causes
```
TypeError: Cannot read property 'split' of undefined
```

jenweber:
Tracked it down to `app/routes/project-version.js`. Making a decision to send the user to the root page in cases that they switch between 2.16 and prior. This is a patch/tech debt. Ideally what should happen here is to do a lookup on some json file/object that contains the past class name mapped to whatever we end up doing for the modules url.

The url is case sensitive, so anything else is too fragile.

jenweber:
### Tech Debt
- [ ] Really, instead of going to the home screen for 2.16 version switching, it ought to map to its equivalent class. Need a reference doc in order to do that.
- [ ] Should write another test for version switching

sivakumar-kailasam:
@jenweber looks like most things in this issue are resolved now. Can we close this?

jenweber:
Yep, remaining issues are captured elsewhere

PR #351, opened 2017-10-05
Patching version switching for 2.16+

jenweber:
This PR preserves version switching for pre 2.15 (i.e. if a user is on Ember.Component for 1.13 and they switch to 2.0, they will still be on the a component page). For v2.16 or higher, users will be brought to the home page when trying to switch versions. This is a patch. Having a classes to modules map would allow for a better solution.

Tech debt is added to https://github.com/ember-learn/ember-api-docs/issues/350


toddjordan:
merged into 2.16 branch

PR #371, opened 2017-10-12
Use "qunit-dom" plugin

Turbo87:
/cc @locks


locks:
@Turbo87 conference-driven development? 😉

Issue #355, opened 2017-10-06
Get the Guides links to work with post 2.16 package restructure


jenweber:
Deadline: released together with 2.16 on Monday

Right now, legacy URLs (like from the Guides) pass through a fancy function that turns them into the right kind of links for the ember-api-docs. We need to modify that function to support the 2.16 packages. 

So there are 2 parts that need to be addressed. One is, how do we catch that it needs to be parsed (so checking if the version is 2.16 or higher for the link) and then mapping it to the right package. Todd is finishing up the final details about how the packages are documented, which influences the links, but I think we could scaffold it out without knowing the final details. Todd will eventually have a table of info to convert from old classes to new modules

New format of sample URL, WIP `http://localhost:4200/ember/2.16/classes/@ember%2Fcomponent%2Ftext-area`

Old format:
`http://localhost:4200/ember/2.11/classes/Ember.Component`

WIP screenshot
<img width="1240" alt="screen shot 2017-10-05 at 11 09 20 pm" src="https://user-images.githubusercontent.com/16627268/31261537-da45bf88-aa22-11e7-977e-35080c49ec24.png">

### Todos

- [x] Make legacy redirects `/classes/Ember.Component.html` default to the latest pre-176 release (2.15.3) #376 
- [ ] Add better error handling for bad urls.
- [x] Support legacy links with element hashes, such as `/classes/Ember.Component.html#method_didReceiveAttrs`
- [ ] Map legacy classes/modules to new 176 counterparts using https://github.com/ember-cli/ember-rfc176-data/ and change redirect version back to the true current release.

toddjordan:
So, 2.16's been deployed, but this is still an open issue.  

#### Links that work:

* Links to previous releases, such as: https://emberjs.com/api/ember/2.11/classes/Ember.Component

#### Links that do not yet work

* Links to unspecified releases, such as:  https://emberjs.com/api/ember/release/classes/Ember.Component
* Legacy links from the old api site, such as http://emberjs.com/api/classes/Ember.Route.html

jenweber:
Missing maps:
- `on()`
- http://emberjs.com/api/classes/Ember.Debug.html#method_registerDeprecationHandler
- http://emberjs.com/api/classes/Ember.Object.html#method_incrementProperty
- http://emberjs.com/api/classes/Ember.Enumerable.html

`Ember.Computed.methodName` references throughout guides should be updated. At risk of perceived churn.

[...]

PR #376, opened 2017-10-18
Fix legacy redirects

toddjordan:
Fixes part of #355 

oddly enough, this doesn't look like it was broke by 2.16.  At some point we changed to fetching project revisions to determine fingerprinted releases.  these routes weren't updated for that new sequence, so all i'm doing here is adding the followup fetch of project-version rev file.

Currently hard coding to 2.15.3 until we can add mapping code.


toddjordan:
Want to write tests and refactor, but also would like to get it out quickly...  if we decide to merge this I'll create a followup issue.

PR #390, opened 2017-10-23
version switching fixes

toddjordan:


app/routes/project-version.js
```diff
@@ -68,17 +68,19 @@ export default Route.extend({
           endingRoute = `namespaces/${namespaceName}`;
           break;
         }
+        case 'project-version.classes.class.methods.method': {
```
toddjordan: actually need to add 2 more cases for events in and properties.

jenweber: I'm also seeing that being on the methods tab blocks route changes. Example: http://localhost:4200/ember/2.14/classes/Ember.Route/methods

toddjordan: @jenweber unsure what you are referring to: 
https://emberjs.com/api/ember/2.15/classes/Ember.Component/methods
https://emberjs.com/api/ember/2.14/classes/Ember.Route/methods

toddjordan: ok. maybe I do. Do you mean if I am on https://emberjs.com/api/ember/2.14/classes/Ember.Route/methods and switch versions, it resets to application module and not to the methods route of the new version?  Fixing that as well.

jenweber: @toddjordan I mean that when I was on the methods tab, if I tried to change versions, the version route change never happened at all. The dropdown and url indicated that I was still on the same version. However version switching worked when I was on the index tab.

toddjordan:
Ok. Should be fixed by the last update. I’m planning on adding some tests
as well

toddjordan:
@jenweber fixed up version switching and added tests

Issue #375, opened 2017-10-17
Make sure `sortedUniqueItems` is being calculated properly


locks:
The logic at https://github.com/ember-learn/ember-api-docs/blob/master/app/components/api-index-filter.js#L41 is a bit suspicious, as `uniq` does not receive additional parameters, and `sortBy` doesn't have quite that signature, either.

From: https://github.com/ember-learn/ember-api-docs/issues/343#issuecomment-337208263

Gaurav0:
May I just simply recommend https://github.com/lodash/lodash/blob/87730eca06d81a8dcc18adbb62eb1a8fcc8e15b3/sortedUniqBy.js

jenweber:
closed with https://github.com/ember-learn/ember-api-docs/pull/408'

Thanks all for your help!

PR #408, opened 2017-12-19
Simplify duplicate filtering

toddjordan:
Addresses https://github.com/ember-learn/ember-api-docs/issues/375

- uses lodash according to documentation.  
- Removed an unnecessary uniq call
- adds a duplicates test case

### DIFF
--- a/tests/integration/components/api-index-filter-test.js
+++ b/tests/integration/components/api-index-filter-test.js
@@ -15,10 +15,20 @@ const model = EmberObject.create({
     version: '2.9'
   },
   name: 'hai',
+  file: 'my-class',
+  parentClass: {
+    file: 'my-class'
+  },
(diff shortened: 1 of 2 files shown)


tests/integration/components/api-index-filter-test.js
```diff
@@ -15,10 +15,20 @@ const model = EmberObject.create({
     version: '2.9'
   },
   name: 'hai',
+  file: 'my-class',
+  parentClass: {
+    file: 'my-class'
```
toddjordan: need to make this my-parent-class

jenweber:
Merging. If there are any issues, they would show up as "new" duplicates

jenweber:
P.S. thanks Todd 😀

Issue #360, opened 2017-10-07
Show examples of import for each class/function


toddjordan:
Each class/function should show how to import/use it.

toddjordan:
My thought on implementing this is to use https://github.com/ember-cli/ember-rfc176-data/blob/master/mappings.json to programmatically generate the import example for each class/function on their specific doc page.

toddjordan:
fixed!

PR #413, opened 2017-12-31
Show import statements for items

toddjordan:
Addresses #360 

uses https://github.com/ember-cli/ember-rfc176-data/blob/master/mappings.json to determine whether to show import help

Example for Class:
image

Example for function:
image


jenweber:
@toddjordan this is so so nice, thank you

Issue #407, opened 2017-12-19
support release links


toddjordan:
Instead of referring to links with specific versions, we should support a `release` link, which would redirect to the current release.  This previously was supported, but has since regressed.

example: https://emberjs.com/api/ember/release/classes/Application

juventinoromero:
Hello @toddjordan, I'd like to help with this.

toddjordan:
hi @juventinoromero thanks!  I've got an in progress fix.  What needs to happen now is I need to get the tests working and create new tests that test the new url format.  Feel free to add to the pr, or add the new tests in a new pr

#409

juventinoromero:
@toddjordan ok! I'll see what I can do. Thanks!

PR #414, opened 2018-01-03
add link rel canonical for release url

toddjordan:
Initial pass at adding link canonical tag to api docs in support of `release` url
Part of addressing #407

### DIFF
--- a/app/templates/head.hbs
+++ b/app/templates/head.hbs
@@ -1,7 +1,6 @@
 <title>{{model.title}}</title>
 
-<link rel="dns-prefetch" href="{{cdnDomain}}" >
-
+<link rel="dns-prefetch" href="{{model.cdnDomain}}" >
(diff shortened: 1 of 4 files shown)


app/templates/head.hbs
```diff
@@ -1,7 +1,6 @@
 <title>{{model.title}}</title>
 
 <link rel="dns-prefetch" href="{{cdnDomain}}" >
```
toddjordan: actually this may need a bug from before, and need to be model.cdnDomain.  will check and update if necessary.

sivakumar-kailasam: Yeah, what u said 😄

sivakumar-kailasam:
Can we add a test for this?

toddjordan:
yeah, will see what I can do for a test.  Should be able to assert on it with an acceptance test

PR #416, opened 2018-01-05
refer to config'd host instead of relative url

toddjordan:
when we initially deployed to prod the code below wasn't fetching the mappings file bc fastly didn't have it.  When it failed we were getting errors.

Therefore, doing 2 things:

* ~~fetching based on the resolved host (instead of relative)~~
* fetch using the routerRootURL (which knows when to add `/api/`)
* on error setting mappings empty, which will just result in the import examples not showing up and let the user proceed, just with out import examples, which is much more graceful.

### DIFF
--- a/app/services/legacy-module-mappings.js
+++ b/app/services/legacy-module-mappings.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import fetch from 'fetch';
 import { task } from 'ember-concurrency';
+import ENV from 'ember-api-docs/config/environment';
 
 export default Ember.Service.extend({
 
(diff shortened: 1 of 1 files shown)


app/services/legacy-module-mappings.js
```diff
@@ -9,9 +10,13 @@ export default Ember.Service.extend({
   },
 
   initMappings: task(function * () {
-    let response = yield fetch('/assets/mappings.json');
-    let mappings = yield response.json();
-    this.set('mappings', mappings);
+    try {
+      let response = yield fetch(`${ENV.API_HOST}/assets/mappings.json`);
```
toddjordan: @sivakumar-kailasam should I actually be using `ENV.routerRootURL` instead?  Maybe that's my problem.  I noticed other assets using it.

```javascript
  if (environment === 'production') {

    /**
     * Ideally we want this to be only for fast boot. But we have to wait for
     * https://github.com/ember-fastboot/ember-cli-fastboot/issues/254 to be
     * solved for that
     */
    ENV.routerRootURL = process.env.DOCS_SLUG ? process.env.DOCS_SLUG : '/api/';
    ENV.gaTrackingId = 'UA-27675533-1';

  }
```

toddjordan: I'm going to go ahead and change it to that.  It makes sense that's what it is...

PR #426, opened 2018-01-12
feat: bring back since value

### LINKED ISSUE #425: Bring back "since" data
This data was available in the old API Docs, but has been removed in this new version. I think this is a very important feature for new and veteran Ember developers alike.

knownasilya:
Don't merge yet, since the `since` seems to be available on every class and field and points to the current version.

Also adds it to each field.

Resolves #425


toddjordan:
hmm. then there must be an additional fix to be had in ember-jsonapi-docs

locks:
Not sure I understand what the root problem is, can someone tldr?

toddjordan:
so I have a PR over at jsonapi docs which seems to enable this (by killing code lol) https://github.com/ember-learn/ember-jsonapi-docs/pull/61

Now the hard part is how to deploy this.  Since all the json docs out there have the "since"s all populated with current version, we would need to rerun the jsonapi-docs script on every release.

We should be able to do this I think, but I certainly am afraid.  @sivakumar-kailasam any thoughts?  I'd like to overwrite all the jsonapi out on s3 with the newly generated files.

toddjordan:
It think I could run the deploy from heroku on each existing version/project combination with the `--ignorePreviouslyIndexedVersion` flag on.  That would force it to rerun, question is will the aws copy overwrite?

sivakumar-kailasam:
All our generated files are fingerprinted, so regenerating them shouldn't be an issue. The only caveat could be that, now that we're reprocessing every version ever by cleaning existing docs, the app would be worked in the meantime. It used to take more than 45 minutes to do this a year ago and now we have more versions out there. 

I propose we pause the script on heroku, run it locally on one of our machines, then once the processing is done, we can copy over the content to s3 with different target folder names(to again avoid borked clients) and then finally rename the old and new folders to swap them in prod. 

I can do these steps if it helps 🙂

toddjordan:
Thanks for the input, @sivakumar-kailasam.  I've run the new "clean" feature on jsonapi docs and it seems to be working well.  I'm unsure how to work with s3 outside the javascript code in the jsonapi docs repo.  I don't know if theres some kind of scp command you can manually move them with.  I can update the script to add a mass local copy option, or if you have a better way I'd be fine with you doing it or pointing me at the appropriate steps.  the jsonapi side is good to go.

sivakumar-kailasam:
I use the s3-cli package from npm for manual interactions with s3.

toddjordan:
Alright, json-docs redeployed, verified fix on heroku preview.  Ready for checkin

PR #436, opened 2018-02-04
create DEPLOYING.md guide

jenweber:
Please review for accuracy and whether it's the right level of scary. Feel free to add comments to request changes or edit directly.

### DIFF
--- a/DEPLOYING.md
+++ b/DEPLOYING.md
@@ -18,13 +18,12 @@ branch," and select the branch you wish to deploy to staging.
 
 ### Deploying to production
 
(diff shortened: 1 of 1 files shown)


sivakumar-kailasam:
@jenweber may be consider adding the reason why we don't recommend promote to production via the button?
> We shouldn't promote from staging with the promote button because heroku uses the same build from staging in the production container, this would mean that the variables set during ember builds get reused. Some of these variables have different values between staging and production which leads to issues when promoting via above mentioned method.

jenweber:
@sivakumar-kailasam edited! Thanks for the insights!

Issue #424, opened 2018-01-11
Functions should get their own page


toddjordan:
Functions should have their own page.  Right now they are grouped by import using the class template.

This is a bit confusing, plus funtions on their own page paves the way to tree navigation (see #344)

PR #440, opened 2018-02-08
Make a route that goes to a specific function

toddjordan:
Addresses #424

Later I'll need to do some work around giving context of module its in, as well as indicating something on the left nav, but its a start and a good intermediate state, as it reduces confusion around loading functions in these class-like page structures. 

This a precursor to getting a collapsable left nav

- [x] Create function route and page
- [x] Update link on module page
- [x] Update link on search result

image


sivakumar-kailasam:
Is there anything we need to do to make this work with search?

Issue #427, opened 2018-01-19
emberjs.com/api does not return fastboot-rendered page


toddjordan:
When you visit a specific API doc URL on emberjs.com, like https://emberjs.com/api/ember/2.18/classes/Engine, the html page is not populated as you would expect fastboot to do.  What we would expect is the site to initially return the completely hydrated page, but what happens is only the initial bare index html is returned, and then hydrated by javascript, as a non-fastboot app would be.  This raises some potential problems for seo/web crawlers, and negates the benefit of fastboot.

``` html
<!DOCTYPE html>
<html lang="en">
  <head>

    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="description" content="">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <link rel="shortcut icon" href="https://ember-api-docs-frontend.global.ssl.fastly.net/assets/images/favicon-f2d07120d2bb7f0d4a63810011903d75.png">

    
[...]

PR #442, opened 2018-02-08
[Dont-Merge] initial naive implementation of prember as a POC

mansona:
Howdy folks 👋 

I started a conversation with bupkizz on the ember community slack about SEO and I didn't know that the fastboot situation was giving you grief. I mentioned that we have had a very successful time of using prember on the new deprecations-app and the new guides-app so I thought I'd start the conversation over here.

You can pull this branch and run `ember build -e production` to see the prember in action, and then have a look at the dist folder to see the results.

screen shot 2018-02-08 at 20 54 00

screen shot 2018-02-08 at 20 54 51

now, this could cause some issues with trailing slashes but I have a way to fix that too 😂 I thought this was at least enough to get the conversation started anyway.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -4,6 +4,8 @@
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 const Funnel = require('broccoli-funnel');
 const mergeTrees  = require('broccoli-merge-trees');
+const fetch = require('node-fetch')
+const _ = require('lodash');
 
 module.exports = function(defaults) {
   let prepend = '';
@@ -12,6 +14,23 @@ module.exports = function(defaults) {
   }
 
   let app = new EmberApp(defaults, {
+    prember: {
+      urls() {
+        return fetch('https://ember-api-docs.global.ssl.fastly.net/rev-index/ember-2.18.0.json')
+          .then(res => res.json())
+          .then(json => {
+            return _.chain(json)
+              .get('data.relationships.classes.data')
+              .map('id')
+              .map(id => {
+                const parts = id.split('-');
+
+                return `${parts[0]}/${parts[1]}/classes/${parts[2]}`;
+              })
+              .value();
+          });
+      }
+    },
     fingerprint: {
       extensions: ['js', 'css', 'jpg', 'png', 'gif', 'map', 'svg', 'webmanifest'],
       prepend,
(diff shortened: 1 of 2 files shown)


mansona:
May fix https://github.com/ember-learn/ember-api-docs/issues/427

sivakumar-kailasam:
I'm actually on the fence about prerendering on the official site. It sends out mixed signals to the community that if not for prerendering fastboot might not be SEO friendly. We did have a brief period where SEO wasn't broken even with fastboot. I feel we should try to spend some time figuring out what could be wrong there before go down this path.

sivakumar-kailasam:
@MartinMalinda 💯 on using right tool for the job. The api docs app is meant to be data driven so that whenever there's a new version of docs released, the app would start showing them without the need for a rebuild/deploy. We could potentially decide at this point that we'll make changes on CI+CD to enable rebuilds & deploy as each version is released. If we're in favor of that then prerendering is probably the way to go.

mansona:
@sivakumar-kailasam one thing that did "fall out" of the implementation when doing pre-rendering on the guides-app and deprecations-app was this business of the trailing slash in URLs. I notice that #427 discusses this a little bit.

Because we needed this concept of folders (with index files) in the guides-app we implemented a [trailing history location handler](https://github.com/ember-learn/guides-app/blob/master/app/locations/trailing-history.js) that always makes sure there is a trailing slash on all urls. This may not be the best thing for your use case but it might help with consistency.

Also for the record, I agree with @MartinMalinda, if your data is 100% knowable ahead of time then we should at lest consider pre-rendering. I have been on this big pre-render exploration with the main motivation being that it's easy to deploy and that is a **good** thing to encourage the community to do.

mansona:
@sivakumar-kailasam out of interest how often are we expecting the data to change?

sivakumar-kailasam:
The index data changes everytime a new version(including patch versions) of ember or ember data is released

mansona:
So if everything is going well it should change once a week (because of the "release train")

I would personally be happy to have a 30-60 mins build time once a week to dramatically simplify the deployment process.

The biggest potential issue is the trailing slash problem that might be a blocker for this implementation

acorncom:
Although we can indeed prerender the api docs, I’d agree with Siva that there is value in having one of our Learn team apps be a true Fastboot app. It provides a reference implementation that we can point folks to (and was part of our original motivation for using Fastboot here in the first place). I’d down the road, we have a better use case for Fastboot in something else, I’d have no problem with switching things around. But not sure I see a benefit in altering what is already working here ...

sivakumar-kailasam:
Thanks for taking the time to try this @mansona. I'm closing this since todd fixed the fastboot issue & the server renders full html now.

PR #443, opened 2018-02-09
Start removing jQuery

rwwagner90:
I started removing jQuery for this app, but then realized that we are using ember-percy here and it still requires jQuery for now, so figured I should PR in the small progress I have and we can finish the rest when ember-percy is updated.

### DIFF
--- a/tests/acceptance/switch-project-test.js
+++ b/tests/acceptance/switch-project-test.js
@@ -6,7 +6,6 @@ import {
   findAll,
   findWithAssert
 } from 'ember-native-dom-helpers';
-import $ from 'jquery';
 
 moduleForAcceptance('Acceptance | Switch Project');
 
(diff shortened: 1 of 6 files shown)


tests/acceptance/switch-project-test.js
```diff
@@ -22,13 +21,13 @@ test('Can switch projects back and forth', async function(assert) {
 
   await click('.spec-ember-data');
   await ensureVersionsExist();
-  assert.ok($(findWithAssert('.spec-ember-data')).hasClass('active'));
+  assert.ok(findWithAssert('.spec-ember-data').classList.contains('active'));
```
MartinMalinda: it's also possible to use `qunit-dom` the syntax will be more similar to the previous jQuery syntax, maybe more readable

rwwagner90: I think `classList.contains` is just as readable as `hasClass`, but that's just my opinion. Happy to try out something else if that is desired.

PR #454, opened 2018-02-17
Landing page for API docs

jenweber:
We have an API docs landing page ready for review!

screen shot 2018-02-17 at 12 47 23 am
screen shot 2018-02-17 at 12 47 34 am
screen shot 2018-02-17 at 12 47 45 am

@jaredgalanis and I would like signoffs from the following people before merging. We expect to make revisions:
- [x] @sivakumar-kailasam or @toddjordan to make sure our routing/redirect edits are sensible
- [x] One or two other team members for a review of content
- [x] one final check by @jaredgalanis after iteration


tests/acceptance/redirects-test.js
```diff
@@ -8,7 +8,7 @@ test('visiting /', async function(assert) {
   await visit('/');
```
toddjordan: Also assert that ember release goes to the landing page.  (pick something pretty stable to assert, like the heading)

ember-cli-build.js
```diff
@@ -36,6 +36,11 @@ module.exports = function(defaults) {
     },
     svgJar: {
       sourceDirs: ['public/assets/images']
+    },
+    'ember-prism': {
```
toddjordan: Still hasn't been explained what prism is giving us, and why it was tacked onto this pr.  That was the screenshots I was asking for.

jaredgalanis: @toddjordan. Prism is the code snippet highlighter library that's being used in the forthcoming guides/deprecations apps. Sorry, I thought I pasted a screen shot in the other issue thread in response to your question, but maybe I didn't get everything.

toddjordan: ok, make sure we check the api code snippets and make sure this doesn't affect them.

tests/acceptance/redirects-test.js
jenweber: Good suggestion, thanks! I added that test coverage.

toddjordan:
Does prism highlight differently than coderay?

jenweber:
@toddjordan what do you mean by highlight?

I'm going through to search for any possible css clashes.

jenweber:
@toddjordan I'll chat with @jaredgalanis about maybe replacing these with some copy/paste coderay elements, in order to avoid CSS clashes. I don't see any, but just looking for them makes me think it's not worth the risk. I'd rather have some ugly html.

toddjordan:
I'm not opposed to replacing our code highlighting scheme, but maybe treat
it as a separate pr and do it globally.

jenweber:
@toddjordan the latest work by Jared removes prism for now

app/controllers/project-version/index.js
```diff
@@ -0,0 +1,13 @@
+import Controller from '@ember/controller';
+import Ember from 'ember';
+const { computed } = Ember;
+
+export default Controller.extend({
+  oldPackageImportSyntax: computed(function() {
+    return Ember.String.htmlSafe("<div class='highlight javascript'> <div class='ribbon'></div><div class='scroller'> <table class='CodeRay'> <tbody> <tr> <td class='line-numbers'> <pre>1\n2\n3\n4</pre></td><td class='code'><pre><span class='keyword'>import</span> Ember <span class='keyword'>from</span> <span class='string'>'ember'</span>;\n<span class='keyword'>export</span> <span class='keyword'>default</span> Component.extend({<span class='comment'>\n  // this is the old way</span>\n});</pre> </td></tr></tbody> </table> </div></div>")
```
toddjordan: out of curiosity, what made you decide to put the markup in a string as opposed to a template?  Also why a computed if the value never changes?

jaredgalanis: @toddjordan the way the markup needed to look to make it format correctly for coderay was strange [see here](https://github.com/ember-learn/ember-api-docs/pull/454/commits/719eef82128257a19beb90d910757bc368bf3328). This caused the linter to throw some errors. The final approach seemed closer to receiving the markup as html string in a model's attribute the way the rest of the API docs seem to handle code snippets.

toddjordan:
I can only get this to work going to ember/release
can't switch versions or go to the landing page with ember/3.0, or ember/2.17

sivakumar-kailasam:
@jaredgalanis The page's title is incomplete. Check the classes route to see how its being handled there.

jenweber:
@toddjordan is that the UX you expect, to be able to see it for earlier versions? Should the links to common docs reflect the release or the selected version? Do you think we will want to change this page in a breaking way in the future? I went back and forth about whether it should show up only for release. Right now that's the intentional behavior but I am not sure of the right answer.

jenweber:
Do not merge - fixing a bug with "home page" links after switching to Ember Data. I'm going to remove the landing page when someone switches to ED. In some future work, it would be nice to have a landing page for ED too.

PR #457, opened 2018-02-18
Updates code snippets on landing page to use CodeRay instead of prism and removes Prism

jaredgalanis:


jenweber:
Merging into a feature branch

PR #458, opened 2018-02-20
Makes import syntax CP’s just regular props. Fixes old package import syntax. Updates title to handle no entity in tokens.

jaredgalanis:
Addresses some of the outstanding review comments in #454.


jenweber:
Merge into feature branch

Issue #422, opened 2018-01-11
Implement a "scroll to TOC" for method/event documentation


locks:
Some classes have extensive documentation, and sometimes after clicking to a method of the class, you'd like to go back to the table of contents to see some other part of the documentation, but your scrollbar is so tiny that you feel disheartened—I'm looking at you https://emberjs.com/api/ember/2.18/classes/Component.

The idea, then, would to introduce some way to "scroll to top" when you're for example at https://emberjs.com/api/ember/2.18/classes/Component/properties/isDestroying?anchor=isDestroying.

kennethlarsen:
👋 @deanpapastrat - Yes, absolutely! I think the main idea here is that we would like a solution that uses native dom apis instead of jQuery. 

Feel free to ping me here or on Discord (same username) if you need help getting started ✨

deanpapastrat:
Awesome! Any design principles we have documented that we can reference? Or should I just make it up as I go

kennethlarsen:
@deanpapastrat So far there's https://ember-styleguide.netlify.com/ as a reference

PR #460, opened 2018-02-26
implemented scroll to top (#422)

HemangiS:
implemented `scroll to top` functionality using button element


toddjordan:
@HemangiS Thanks for this PR!  We are trying to wean ourselves off of jquery, so optimally we'd like a solution that uses native dom apis to do the scroll to top.  See https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

PR #484, opened 2018-04-08
[wip]landing page on index

toddjordan:
currently the landing page is on /api/ember/release/
this fix puts it on /api/

this is to give an overal entry point to docs (regardless of project), and to workaround a fastboot bug https://github.com/ember-fastboot/ember-cli-fastboot/issues/349


jenweber:
@toddjordan can we still link to the "normal" landing page from the Guides and the Navbar? This page just exists for SEO purposes, right? Is it not possible for us to have a navbar? Could we redirect to the regular page?

I'm just wondering what the story is for how someone gets on this page, and whether the "missing" info is by necessity. I totally support making the changes if it helps our SEO.

We should make the main body of the landing page a component so that it only has to be maintained in 1 place. I like the logos idea.

toddjordan:
@jenweber I linked to the fastboot issue we are having a problem with.  We can't redirect atm because that is the whole thing with the issue.

Eventually we want the main landing page at the root to probably be different than the ember one.  Eventually we want api docs to encompass our main ecosystem docs including CLI, testing, engines, fastboot, etc, and would include much more general data.

Perhaps we could change the url from emberjs.com to go straight to api/ember/release.   We can even do that without this PR.  What do you think?

cc @locks

toddjordan:
Went ahead and created a PR for the easy thing: https://github.com/emberjs/website/pull/3271

We can leave this wip until we get a good home page concept that separate than what we have for ember/release.  (maybe when we add more projects)

toddjordan:
shelving for now

Issue #491, opened 2018-04-14
Canonical URL broken for fastboot in prod


williamhaley:
I noticed this in prod, and can re-create locally.

![prod canonical broken](https://user-images.githubusercontent.com/1851748/38764074-bc88afe2-3f6d-11e8-8c31-1e7bf1eac10d.png)

![canonical broken](https://user-images.githubusercontent.com/1851748/38764043-5e9497e8-3f6d-11e8-932f-5fe3d5cbc5df.png)

One can see the issue for any [non-release docs](view-source:https://emberjs.com/api/ember/2.18/classes/ArrayProxy). 

```
<link rel="canonical" href="http://[object Object]/ember/release/classes/ArrayProxy">
```

Seems like the issue is that fastboot's `request.host` is a computed property, and not a string.

I'm getting a PR together now.

williamhaley:
@toddjordan Sorry, it looks like I only addressed part of the issue :-/

 Looks like the host is now `http://ember-api-docs.herokuapp.com/ember/release/modules/@ember%2Fobject`

Not sure how fastboot infers what the host should be. Maybe it's just better to hardcode to `emberjs.com` when using fastboot?

Or configure a whitelist like here? https://github.com/ember-fastboot/ember-cli-fastboot#host

toddjordan:
yeah, I think for now a hardcode is fine, since we always want folks to come through emberjs.com/api

williamhaley:
Will do :+1:

williamhaley:
@toddjordan done

toddjordan:
Thanks @williamhaley !  looks good.  This solves our problem for 2.16+,  now we need a function that converts pre-2.16 stuff to current (via https://github.com/ember-cli/ember-rfc176-data/blob/master/mappings.json) . Leaving this open until we get there.

williamhaley:
@toddjordan Is there anything I can do to help with the pre-2.16 stuff? Seems like the Google voodoo is still kinda bad right now. A search for "Ember computed" yields the following. `2.15` docs are at the top with their broken `canonical` url. I'd love to help but I'm not sure where's best to look. No relevant recent docs show up (at all, as far as I can tell) in Google.

_Also, as an aside, the docs hosted at `www.emberjs.com.cn` seem to steal a lot of SEO away from "official" Ember docs. Does the Ember org own that? Is that a static mirror that can be shut down?_

[...]

PR #492, opened 2018-04-14
Fix issue with fastboot canonical URL for API docs

williamhaley:
Fix issue #491. The way we infer the canonical URL for fastboot seems to be inaccurate. The issue exists in prod and I was able to verify it locally too.


williamhaley:
I'd love to add a test for this, but I don't know how to write an acceptance test with/without fastboot (never worked with fastboot before). 

If there's a way to test that please let me know how and I can add something.

PR #496, opened 2018-04-19
Add heading anchors to allow deep-linking

### LINKED ISSUE #494: Allow deep-linking to package method lists
One thing that I've noticed with the recent API docs update is that I've not been able to deep-link to function lists of packages, most notably when linking to Ember's Computed Property Macros list.

I think it would be useful to add IDs to the method list headings and maybe 🔗 icons on hover for easy copying of links.

screen shot 2018-04-19 at 10 19 10

kevinansfield:
closes #494
- adds `id` attributes to headings so they can be linked to with #anchor URLs
- utilises the same "anchorable" markup and image as used in the Ember guides to show a 🔗 icon next to headings on hover


toddjordan:
Tried this approach out on https://ead-review-app-pr-496.herokuapp.com/api/ember/release and it seems to be failing.  I click on one of the sub module deep links, copy the url, then paste in another tab.  The page loads, and then refreshes to a 404.

toddjordan:
Validated with recent build and the issue I was seeing before is resolved (since we removed some of the legacy redirect code).  Went ahead and push up your changes manually to master
image
Therefore closing

PR #501, opened 2018-04-23
[WIP] 3.x upgrade

sivakumar-kailasam:
- Ran ember-cli-update with codemods
- Simplified test assertions with qunit-dom
- Introduced prettier setup that'll work as code is git staged on local
- Replaced hardcoded config value with actual reference to config.


toddjordan:
First off thanks for this (its a lot!).  Tried to resolve conflicts on but finding it hard, partially because of its age.  I also think its trying to do a bit too much for one pr (prettier, replacing libs with custom versions, etc), which I'd like to handle more incrementally.  I'd like to take a bit more incremental approach.  I've started the process in PR #570

sivakumar-kailasam:
No worries, I was a bit ambitious when I started on this 😅. Your approach is much more realistic.

PR #508, opened 2018-05-07
Convert legacy urls to "release" route instead of 2.15

toddjordan:


tests/acceptance/convert-legacy-url-to-current-test.js
```diff
@@ -0,0 +1,45 @@
...
+});
+
+test('should convert unknown legacy modules url to landing page', async function (assert) {
+  await visit('/modules/ember-metal');//no .html
+  assert.equal(currentURL(), '/ember/release');
+});
+
+test('should convert unknown legacy data url to 404', async function (assert) {
+  await visit('/data/modules/ember-random.html');
+  assert.equal(currentURL(), '/404');
```
toddjordan: The reason we go 404 for ember data urls that we can't map is that none of the ember data classes are mapped in json (yet), therefore we have to just try to load the conversion, which is `/data/classes/<classname>.html` to `/ember-data/release/classes/<classname>`.

toddjordan:
Don't merge yet.. Checking this out on https://ead-review-app-pr-508.herokuapp.com/api/ember/release/classes/Component

While it works from a user perspective, I'm seeing an error from fastboot, which returns an error page.  User doesn't see it because it gets overridden by client side javascript, but will probably screw up search crawlers.

Looking into it now.

toddjordan:
Ok, could be because we fetch from the fastly cdn when we get the mappings and the review apps don't have it set.  Trying to see if I can get this installed on one of the fastly enabled apps.

toddjordan:
Alright, added something to avoid the error to environment.js.  @sivakumar-kailasam could you look this over?  Open to better approaches.

config/environment.js
```diff
@@ -24,11 +24,11 @@ module.exports = function(environment) {
       // Here you can pass flags/options to your application instance
       // when it is created
       scrollContainerSelector: 'body, html',
-      cdnUrl: process.env.FASTLY_CDN_URL ? `https://${process.env.FASTLY_CDN_URL}` : ''
+      cdnUrl: process.env.FASTLY_CDN_URL ? `https://${process.env.FASTLY_CDN_URL}` : 'https://ember-api-docs-frontend.global.ssl.fastly.net'
```
toddjordan: fetch throws an error on fastboot if you don't have an absolute url. This was causing problems in local envs and the review app.  Downside to this is that dev envs are making requests to fastly.  I don't think this should be a deal, but optimally would like it to be localhost when running locally.  Maybe we can redefine this variable in the `if (environment === 'development') block.

PR #521, opened 2018-06-09
Application content container size fix on empty content

lifeart:
Before:
image

After:
image

(https://www.emberjs.com/api/ember-data/3.2/modules/ember-data)


lifeart:
Yeah, looks like browserstack CI is broken

Issue #47, opened 2016-06-01
opengraph tags for methods/events/properties should have more information


fivetanley:
right now they don't show anything

![](https://dl.dropboxusercontent.com/s/o4vs3j069nca8ct/2016-06-01%20at%2011.09%20AM.png)

mschinis:
picking this one up, I'm going to add additional meta tags using ember-cli-meta-tags.

oskarrough:
@mschinis did you get going or maybe I can help?

I'd also like to remove the tomster open graph image. It takes up a lot of unnecessary vertical space as links are shared. We could replace with a little description, if that's useful.

toddjordan:
@oskarrough go for it :-)

oskarrough:
Alright! I looked at it and it seems meta tags are already setup, just a few missing things.

Are you OK with removing the tomster, too? Would be great to have a Slack chat with useful descriptions instead of the logo over and over again.

sivakumar-kailasam:
I find the guides urls to be descriptive on slack. Can you please use that to cross verify? The logos themselves might not be the root cause for the lacking experience

oskarrough:
Hey @sivakumar-kailasam, what do you mean by cross verifying the URLs? :)

I think what we have to do is 

1) make sure there are meta descriptions where possible
2) remove huge tomster logo

I agree that the logo isn't the root cause but it will nicer to share API links without it as it wastes 300px every time…

I noticed that the API returns descriptions filled with HTML. I'll make a helper to strip it out or do we have clean descriptions without markup? Example here in `data.attributes.description` https://ember-api-docs.global.ssl.fastly.net/json-docs/ember/3.2.0/classes/ember-3.2.0-ComputedProperty-a0c6b9793e.json

sivakumar-kailasam:
I meant cross verify the  meta tag descriptors used on both pages :) You can probably add the helper to extract the text description.

PR #534, opened 2018-06-27
Improve meta tags

oskarrough:
This PR improves the meta tags to make API links more useful as they are shared around the web. This includes…

- creating a truncated excerpt from the descriptions
- remove huge og:image to save ~300px each time a link is shared in Slack

Here's an example description for http://localhost:4200/ember/3.2/functions/@ember%2Fdebug/registerDeprecationHandler:

> Allows for runtime registration of handler functions that override the default deprecation behavior. Deprecations are invoked by calls to @ember/application/deprecations/deprecate. The following example demonstrates its usage by registering a handler that throws an error if the message contains the…

For routes that don't provide a description, such as http://localhost:4200/ember/3.2/modules/@ember%2Fobject, what could we do? Maybe we leave those for now. Any other improvements we could do now that this is open?

Closes https://github.com/ember-learn/ember-api-docs/issues/47


oskarrough:
Examples

example

sivakumar-kailasam:
@oskarrough to answer some of your questions, the next possible improvement could be in the source of the docs themselves where it could provide more insights into the API. I'm sure others have ideas on that as well.

Some of the open issues might have those details already

PR #544, opened 2018-08-09
[FIX] Fix issue #543 - update redirect logic for <=2.15 to/from >=2.16

### LINKED ISSUE #543: Don't redirect ember-data lookups to home page on version switch from <=2.15 to >=2.16
Currently, there isn't a way to map ember classes after the changes in 2.16, and the API docs redirect users to the home page of the API docs when they attempt to change from a version 2.15 or lower to a version 2.16 or higher. 

We should exclude ember-data from this redirect as it does not have the same issue relating to changed class names, and so the version switcher should continue to work for ember-data.

localpcguy:
- Allow redirects for <=2.15 to/from >=2.16 if project is ember-data
- Continue to disallow redirects for <=2.15 to/from >=2.16 if project is ember
- Allow landing page to be displayed for versions >=2.16
- Add tests for ember-data version changes

Fixes #543

### DIFF
--- a/app/routes/project-version.js
+++ b/app/routes/project-version.js
@@ -34,10 +34,11 @@ export default Route.extend({
     let transitionVersion = this.get('projectService').getUrlVersion();
     if (!classParams && !moduleParams && !namespaceParams && !functionParams) {
       // if there is no class, module, or namespace specified...
-      let latestVersion = getLastVersion(model.get('project.projectVersions'))
-      let isLatestVersion = (transitionVersion === latestVersion || transitionVersion === 'release')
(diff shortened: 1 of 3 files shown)


localpcguy:
Something introduced an issue with scroll position when transitioning between versions >= 2.16 and <= 2.15; I put in a test for it, but ran outta time to fix it today. Not sure if that should be a new issue or if it needs to be fixed as part of this. It seems to be working properly in the browser, so it seems like a test issue - but maybe someone could verify that?

toddjordan:
Looks great!  Thanks @localpcguy

PR #550, opened 2018-09-17
Remove check for @ember/object/computed to keep it out of contents

toddjordan:
Added a check to keep @ember/object/computed from showing up in TOC until ember source was fixed.  Its since been fixed


locks:
@toddjordan for all relevant versions of Ember?

toddjordan:
Gonna close this PR because while the issue may be fixed in current releases, it will rear its head in some old ones.  Until we correct the issue in older docs deployments we need to keep the workaround, or let the issue show up in those old releases

Issue #519, opened 2018-05-30
Clicking a link to a method from ember guides -> ember api docs jumps to wrong place after render


tchan:
Steps to repro:
- Visit https://guides.emberjs.com/v3.1.0/object-model/computed-properties-and-aggregate-data/
- Click ` computed.filterBy` link

Using latest Chrome and attached gif below. When you click the link it loads to the right place for half a second before it fully loads in which it is no longer in the correct spot and you have to scroll down.

![filterby](https://user-images.githubusercontent.com/18437085/40700066-3080482a-641b-11e8-8660-fd66453b16d7.gif)

toddjordan:
This could be a bug in the link in guides which is: https://www.emberjs.com/api/ember/release/classes/@ember%2Fobject%2Fcomputed/methods/alias?anchor=filterBy , which should be https://www.emberjs.com/api/ember/release/classes/@ember%2Fobject%2Fcomputed/methods/filterBy?anchor=filterBy

A fix should be made to ember-learn/guides-source

locks:
Seems fixed.

PR #558, opened 2018-10-02
Fix anchor scrolling by making sure all data is loaded before render

HuggableSquare:
Related to #519 

The import examples were loading after the page had loaded, pushing everything down and causing the scroll to anchor to be wrong.

One failing test is fixed in #557, I figured it didn't make sense in this PR since it was caused by something completely separate.

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -6,6 +6,7 @@ import getCompactVersion from 'ember-api-docs/utils/get-compact-version';
 
 export default Route.extend({
   headData: service(),
+  legacyModuleMappings: service(),
 
   title(tokens) {
     let [version, entity] = tokens;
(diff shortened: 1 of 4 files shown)


toddjordan:
Thanks for the fix @HuggableSquare. will do some testing on this tomorrow.

toddjordan:
Preview App is here: https://ead-review-app-pr-558.herokuapp.com/api/ember/release
I noticed that import examples don't show up for classes in this preview app, but they do in production.
Will need to figure out why they no longer load.

HuggableSquare:
That was my bad, I accidentally replaced `(not (eq static 1))` with `(eq static 1)` in `app/templates/project-version/classes/class.hbs`, should be fixed now.

PR #560, opened 2018-10-04
Moved css style for the mouse cursor in search results

cah-danmonroe:
Moved the `cursor:pointer;` to the clickable search results content wrapper.  Now only the links that actually do something will appear as "clickable" to the user.


toddjordan:
Thank you!!!!

PR #561, opened 2018-10-05
Stop resetting filtering query params on tab change

### LINKED ISSUE #515: Filter for Public/Private/Inherited/etc... Resets when clicking on method/property
When clicking on a method name the Public/Private/Inherited Filters reset.

This means that if you are trying to view a method that is private or deprecated you get bounced out and cannot view the method/property that you are trying to view.

Steps to recreate

1. Go to any class in the API docs: ex https://www.emberjs.com/api/ember/release/classes/Controller
2. Activate Protected or Deprecated filter: https://www.emberjs.com/api/ember/release/classes/Controller?show=inherited%2Cprivate
3. Click on a method and it will not show: https://www.emberjs.com/api/ember/3.1/classes/Controller/methods/_qpChanged?anchor=_qpChanged

HuggableSquare:
Fixes #515 

I tried to write an acceptance test for this, but I couldn't manage to get one that failed under the old code. This bug is very elusive, and I kind of suspect fastboot might be part of it (although I've never used fastboot so I could be very wrong here).

It seems like sometimes you will load the page and it will work perfectly as intended, and other times the bug will pop up. The best reproductions I've found are either what was said in #515 (opening the url to the class directly in a new tab and then trying the bug), or going to the home page in a fresh tab and then going to the class page. From what I can tell, for some reason in these cases the link-to's never update with new query params on them when the values change, causing the bug.

With that said, this PR has solved the bug in those cases, and I haven't been able to reproduce it with this change, so I figured I would throw it up for y'all to look at and decide what to do. :)

### DIFF
--- a/app/controllers/project-version.js
+++ b/app/controllers/project-version.js
@@ -6,10 +6,9 @@ import { inject as service } from '@ember/service';
 import values from 'npm:lodash.values';
 import groupBy from 'npm:lodash.groupby';
 import semverCompare from 'npm:semver-compare';
-import FilterParams from '../mixins/filter-params';
 import getCompactVersion from '../utils/get-compact-version';
 
-export default Controller.extend(FilterParams, {
+export default Controller.extend({
 
   filterData: service(),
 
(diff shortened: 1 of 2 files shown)


toddjordan:
Thanks!

Issue #525, opened 2018-06-15
Javascript ribbon of examples not displayed properly in the API documentation


HenryVonfire:
**Description**
There are two bugs but the can be related so I add them here:

1. In the code example of the API documentation, the `JS` label overlaps the content displayed
2. The `copy to clipboard` button is always visible for the first code example and never for the rest of the examples. 

**Screenshot**
![ember_screenshot](https://user-images.githubusercontent.com/10612765/41464461-b2ce2a1e-709a-11e8-8c13-3c9f20ba9fd3.png)

**Browsers where it happens**
So far I've tested that it happens in Firefox and Chrome

**Links**
this [isPresent](https://emberjs.com/api/ember/3.2/functions/@ember%2Futils/isPresent) is an example but it happens in every place an example of code is displayed.

**Versions** 
It affects all the versions as far as I know.

**Expected behaviour**
This happens in the guides when the ribbon has a text
![ember_expected](https://user-images.githubusercontent.com/10612765/41464594-2edb3d04-709b-11e8-96a5-3b82debcad27.png)
And this when it hasn't
![ember_expected2](https://user-images.githubusercontent.com/10612765/41464664-6388619e-709b-11e8-8eeb-d79de7d9f316.png)

@mansona gave me more insight about it.
![screenshot](https://user-images.githubusercontent.com/10612765/41464794-d7f20dfa-709b-11e8-8f9e-21a43bebaa03.png)

[guides-article link](https://github.com/ember-learn/guides-app/blob/master/app/components/guides-article.js#L20)

/cc @toddjordan , @sivakumar-kailasam

toddjordan:
Thanks for reporting.   I've noticed that the content certain pages  are compressed horizontally even when viewing on larger screens.  There's a CSS fix to be had for this.   The second issue is that the copy button is inconsistently showing up and misaligned with the code.  the third would be overlapping for one line js examples. We could put a title banner in, though I'd also be for hiding the js banner for those one line import examples.  I don't think its info that the user needs there.  I'd like to keep the copy capability though if at all possible,  its nice when looking up the import for a specific item and putting it in your own code.

PR #562, opened 2018-10-06
Make import example function better on small screens

### LINKED ISSUE #447: inline example slightly off with new copy button
The black strip doesn't extend as far as the other code examples.

Happens at most non-mobile resolutions.  I've only looked on Chrome so far.

image

HuggableSquare:
Fixes #447 and #525

Demo


toddjordan:
This is awesome.  Thanks @HuggableSquare !

PR #565, opened 2018-10-25
Use styleguide navbar to make consistent with the rest of the site

toddjordan:
image


toddjordan:
Some of the spacing of main content got messed up with this. looking at it now...

PR #570, opened 2018-10-26
Upgrade ember to 2.18

toddjordan:
This is an incremental approach to getting api docs up to the latest ember.  I'm using pr #501 as a resource, as it grew and diverged a bit from the latest api-docs because we let it live a bit too long in pending.  Ultimate goal is the same.


config/targets.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  browsers: [
+    'ie 9',
```
toddjordan: Should we support IE 9?

MODULE_REPORT.md
```diff
@@ -0,0 +1,14 @@
+## Module Report
```
sivakumar-kailasam: This is an autogenerated file. We don't need this

config/targets.js
sivakumar-kailasam: No, let's do IE11 & above. The same as ember 3.x support

Issue #532, opened 2018-06-22
Broken Link: Ember.evented (in ember-data/release/classes/DS.RecordArray)


hakilebara:
<img width="1195" alt="image" src="https://user-images.githubusercontent.com/1991564/41787940-2e66479c-764a-11e8-91a1-b0550b0300c3.png">

https://emberjs.com/api/ember-data/3.1/classes/Ember.Evented

sivakumar-kailasam:
@hakilebara this requires a fix in the yuidocs of ember-source. Would you be interested in making a PR for it?

PR #573, opened 2018-10-27
Fix broken uses links

jeffhertzler:
fixes #532

PR #574, opened 2018-10-31
Fix multiple problems with uses and extends

spruce:
As described in the Issue https://github.com/ember-learn/ember-api-docs/issues/532 the links created for the Class [`DS.RecordArray`](https://emberjs.com/api/ember-data/3.5/classes/DS.RecordArray) are wrong. 
Instead linking to 

/api/ember/3.5/classes/**Ember.ArrayProxy** and /api/**ember-data**/3.5/classes/Ember.Evented 
when they should link to 
/api/ember/3.5/classes/**ArrayProxy** and /api/**ember**/3.5/classes/Evented

This is achieved by creating a new helper that removes the `Ember.` prefix and an update to the `projectNameFromClassName()` function.

Remaining problems (not caused by this PR):
- [x] class model only has single value for `usedClassProjectName` where the basis for the creation of that prop is a array with possibly multiple entries
- [x] Some Classes extend from the "wrong" Classes, e.g. `DS.Adapter` uses Ember.Object where it should be EmberObject. I will try and find many errors like this in another PR in ember-data source.

### DIFF
--- a/app/helpers/project-name-from-class-name.js
+++ b/app/helpers/project-name-from-class-name.js
@@ -2,7 +2,7 @@ import { helper } from '@ember/component/helper';
 
 export function projectNameFromClassName([className, fallback]) {
     className = className || "";
-    if (className.indexOf('Ember.') > -1) {
+    if (className.indexOf('Ember') > -1) {
       return 'ember';
     }
 
--- a/app/helpers/project-name-from-class-name.js
+++ b/app/helpers/project-name-from-class-name.js
@@ -1,0 +1,16 @@
+import { helper } from '@ember/component/helper';
+
+export function projectNameFromClassName([className, fallback]) {
+    className = className || "";
+    if (className.indexOf('Ember.') > -1) {
+      return 'ember';
+    }
+
+    if (className.indexOf('DS.') > -1) {
+      return 'ember-data';
+    }
+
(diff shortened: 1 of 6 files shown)


spruce:
As the commits from the PR #573 are now part of this PR this superseeds #573 and can therefor be closed.

toddjordan:
Thanks for looking at this.  I think we should rather fix the reference in the original documentation (emberjs source).  This is really just adding a workaround to the UI.  We can do this by just change `@uses` `@extends` to their current classnames. (without the `Ember.`)

spruce:
Ok. So just that we are on the same page. Ember should not be prefixed and Ember-data will have `DS.` prefixed. I will then go in and fix that in the places where I find it.

toddjordan:
@spruce Yeah, with the onset of JS module import, Many of the classes traditionally referenced using the Ember global namespace are now referenced with imports.  In these cases the ember namespaces was removed from inline documentation.  For things like ArrayProxy, you can just refer to `ArrayProxy` and it will typically resolve it, though not sure about references from ember data (bc it is a separately built documentation)

toddjordan:
Closing in favor of fixing references in the source documentation

spruce:
:+1:

PR #578, opened 2018-11-30
Fixes 404s from search result links for ember-data

cah-danmonroe:


app/components/search-input/dropdown-result.js
```diff
@@ -9,6 +9,12 @@ export default Component.extend({
   groupName: '',
   groupPosition: 0, // Index of this result in the grouped results
 
+  module: computed('result.{project,module}', function () {
+    if (this.get('result.project')) {
+      return this.get('result.project');
+    }
+    return this.get('result.module') === 'ember-data' ? 'ember-data' : 'ember'
```
sivakumar-kailasam: Is this happening only for ember-data module?

cah-danmonroe: As fas I can tell, yes.

PR #584, opened 2018-12-09
Upgrade few deps

sivakumar-kailasam:


toddjordan:
will need to update failing tests before deployment

Issue #596, opened 2019-03-08
Deploy the API docs on api.emberjs.com


mansona:
- [x] Remove the `rootURL` implementation for the API docs so that it doesn't need the `/api/` prefix to all paths
- [x] Deploy the branch with no `rootURL` on Heroku
  - We need to old site to still work while we are still swapping over
- [x] Setup a new fastly service for the api app.
- [x] add DNS entry for api.emberjs.com
- [ ] add redirects from emberjs.com/api to api.emberjs.com

sivakumar-kailasam:
The app's live on https://api.emberjs.com.

PR #597, opened 2019-03-08
Subdomain setup

sivakumar-kailasam:
DO NOT MERGEEEE PLZZZZ! 🙂

implements #596


toddjordan:
@sivakumar-kailasam Looks like api.emberjs.com is deployed.  Should we merge this or close it?

sivakumar-kailasam:
It’s at a point where it’s good to merge. But don’t do it yet but it needs some background chores to be done for it to work after the merge. I’ll do this today & retire the old app.

Issue #598, opened 2019-03-14
Remove potentially unused scroll setup


acorncom:
While thinking about the scrolling system in this app, I noticed that https://github.com/ember-learn/ember-api-docs/blob/master/app/initializers/scroll-parent-shim.js appears to be unused.

It would be quite helpful to have someone double-check me on that and if that is indeed the case, to send in a PR with that removed. Doing so would also help reduce our usage of jQuery in the app, which would help with https://github.com/ember-learn/ember-api-docs/issues/441

rahulguptaio:
I'm going to investigate whether this library can be removed

PR #599, opened 2019-03-18
Removing Scroll Parent Initializer #598

rahulguptaio:
1. Removing the 'scroll parent' initializer and test files to reduce Jquery usage in app. 
#598 

Steps to verify if the function is required -
1. Added a debugger to the scrollParent function in the initializer.
2. Ran the entire test suite to check if the breakpoint is hit.
3. No breakpoint is hit, and no test failures either.


toddjordan:
LGTM, thank you!

PR #603, opened 2019-03-31
 Fix version switching for class names that contains slashes

danwenzel:
This fixes version switching for classes whose name contains slashes, like `@ember/object/computed`

### Reproduction Steps:
1. visit https://api.emberjs.com/ember/3.4/classes/@ember%2Fobject%2Fcomputed/
2. use the version switching dropdown to select version 3.7

Before this fix, you'll notice that no navigation takes place, and you remain on 3.4.  This is because the class's `name` is `@ember/object/computed`, and when the router tries to transition, it treats the slashes as subroutes in the url and can't find a route.

I fixed this by first wrapping the class name in `encodeURIComponent` before transitioning.

### DIFF
--- a/app/routes/project-version.js
+++ b/app/routes/project-version.js
@@ -1,6 +1,5 @@
 import { inject as service } from '@ember/service';
 import Route from '@ember/routing/route';
-import { computed }  from '@ember/object';
 import semverCompare from 'npm:semver-compare';
 import getCompactVersion from 'ember-api-docs/utils/get-compact-version';
 import getFullVersion from 'ember-api-docs/utils/get-full-version';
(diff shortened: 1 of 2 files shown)


app/routes/project-version.js
```diff
@@ -84,18 +85,23 @@ export default Route.extend({
     };
   },
 
+  encodedNameForCurrentClass: computed(function () {
```
danwenzel: Just realized I shouldn't be using a computed property here, as it won't ever update.  Lemme fix this...

danwenzel: Alright, I fixed that, and added a regression test for it

toddjordan:
LGTM... Thank you!!!

Issue #441, opened 2018-02-08
Remove jQuery


rwwagner90:

toddjordan:
Does this issue mean remove usage of jquery in the app, or remove documentation of jquery from the api docs?

rwwagner90:
It means remove usage from the app.

dbendaou:
I could tackle this one

dbendaou:
After some investigation, it seems that `ember-guide` itself as jQuery as an dependencies.
I've started with removing jQuery from tests suits.

sivakumar-kailasam:
jQuery itself still needs to be removed from the app after ensuring that the addons used in the app don't require it.

locks:
I did a quick search and we still seem to be using a jQuery plugin.

PR #610, opened 2019-04-23
Removing jquery from tests

dbendaou:
This PR aims to remove `jQuery` from tests 

Related issue #441

### DIFF
--- a/tests/acceptance/scroll-reset-on-transition-test.js
+++ b/tests/acceptance/scroll-reset-on-transition-test.js
@@ -1,5 +1,4 @@
 import { visit } from '@ember/test-helpers';
-import $ from 'jquery';
 import { module, test } from 'qunit';
 import { setupApplicationTest } from 'ember-qunit';
 import config from 'ember-api-docs/config/environment';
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -41,3 +41,6 @@ local.log
 /.node_modules.ember-try/
 /bower.json.ember-try
 /package.json.ember-try
+
+# Idea
+.idea
```
Turbo87: this should be added to your [global gitignore file](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore)

tests/acceptance/scroll-reset-on-transition-test.js
```diff
@@ -1,76 +1,75 @@
 import { visit } from '@ember/test-helpers';
-import $ from 'jquery';
 import { module, test } from 'qunit';
 import { setupApplicationTest } from 'ember-qunit';
 import config from 'ember-api-docs/config/environment';
 
 const { scrollContainerSelector } = config.APP;
+const scrollSelector = document.querySelector(scrollContainerSelector);
```
Turbo87: it might be better to do this in a `beforeEach()` block to make sure that the DOM doesn't leak between tests

PR #611, opened 2019-04-23
Remove unmaintained ember-algolia and ember-browserify

tchak:


sivakumar-kailasam:
Tested this on my local and it LGTM. So merging it.

PR #647, opened 2019-11-08
Fix broken and unreliable tests on master

Gaurav0:
Tests are failing for me on my local machine without this.

Note that we are seeing `yarn test` return an exit code of zero (and thus a check mark) even though tests are failing in CI.


Gaurav0:
@sivakumar-kailasam Unfortunately no. The actual issue are bugs in old versions of `ember-concurrency` and `@ember/test-helpers`. However, I'd prefer to have tests passing and deprecations fixed before trying to upgrade any of that stuff.

Issue #622, opened 2019-07-05
Broken linkage between ember-data and ember API docs


ef4:
On this page:

https://api.emberjs.com/ember-data/3.10/classes/DS.RecordArray

 - the link with text `Ember.ArrayProxy` is broken.
 - and the methods and properties inherited from `Ember.ArrayProxy` are missing.

toddjordan:
This needs to be changed in the ember data source to just point at `ArrayProxy` instead of `Ember.ArrayProxy` .  We no longer use/doc the Ember Namespace in new releases and the doc links reflect that.

efx:
@toddjordan I generated ember-data documentation locally from master and it appears to have been updated according to your comment:
https://github.com/emberjs/data/blob/282320abe1f73ffe21ca840a002131a8d710b60d/packages/store/addon/-private/system/record-arrays/record-array.js#L21 
That was introduced in https://github.com/emberjs/data/pull/6280.

Locally I see it prints out as `ArrayProxy` but the URL for that link is http://localhost:4200/ember-data/3.15/classes/ArrayProxy:
<img width="1303" alt="Screen Shot 2019-11-15 at 15 56 29" src="https://user-images.githubusercontent.com/472721/68975243-ba5d1880-07c0-11ea-8982-c723de1f1ed5.png">

The target destination doesn't exist as I would expect. If I understand it correctly it should link to `/ember/release/classes/ArrayProxy`?

This also still happens on the current release documentation page for `ArrayProxy`:
https://api.emberjs.com/ember-data/release/classes/RecordArray

efx:
This is a duplicate of #591.

efx:
From a recent search of data, this is in a few places too:
```sh
rg Ember.ArrayProxy
packages/model/addon/-private/errors.js
82:  @extends Ember.ArrayProxy

packages/store/addon/-private/system/promise-proxies.js
39:  @extends Ember.ArrayProxy
129:  @extends Ember.ArrayProxy
```

Gaurav0:
@toddjordan , As @efx mentioned, changing `@extends Ember.ArrayProxy` to `@extends ArrayProxy` in the source does not fix the link. @sivakumar-kailasam I was thinking of working on another redirect solution for this inside ember-api-docs. Do you think this is the best path forward?

sivakumar-kailasam:
[...]

PR #663, opened 2019-11-17
fix links from ember-data to ember

Gaurav0:
This is another redirect to fix links from Ember Data to Ember for `Extends` and `Uses` classes. This does not fix inherited items not showing up.

For #622

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -32,6 +32,7 @@ module.exports = {
         'lib/**/*.js',
         'bin/*',
       ],
+      excludedFiles: ['config/deprecation-workflow.js'],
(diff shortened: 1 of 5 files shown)


Gaurav0:
@sivakumar-kailasam I've already tried it. It requires a url rather than a file path to the data.json file, and fetches it at run time in the generated yuidoc files in the dist directory when you serve them, rather than at build time when you generate it.

MelSumner:
@sivakumar-kailasam please re-review so we can proceed or close. Thank you!

sivakumar-kailasam:
Thanks for your patience @Gaurav0 🙂

Issue #667, opened 2020-02-11
Add Ember CLI Api docs


mehulkar:
Didn't see anything filed already for this, so kicking this off. 

I looked into how this repo works to see if I could port over https://ember-cli.com/api/, but looks like it's making a request to api-store.emberjs.com, which points to AWS. I'm assuming some script/job is building and posting docs at release time to AWS, and the same needs to happen for Ember CLI. Will post more of my findings here as I discover how it works. If there's any existing work for this, let me know!

sivakumar-kailasam:
👋 Hey @mehulkar, thanks for reporting this 🙂. [ember-jsonapi-docs](https://github.com/ember-learn/ember-jsonapi-docs) handles the creation of content that gets published to the aws store. 

A new rewrite is underway and is in its early stages (you can follow the progress [here](https://github.com/ember-learn/ember-api-docs/projects/5)). While we might not provide versioned docs for ember-cli, we'll certainly include it as a part of the new app.

mehulkar:
Awesome. Would it make sense to at least add a page that showed *some* info for now? e.g. #668

PR #668, opened 2020-02-11
Add page for Ember CLI API Docs

mehulkar:
I had already done this before I saw the comment in #667. Figured i'd send it in. Seems like an improvement even if there is something better coming.

### DIFF
--- a/app/routes/ember-cli.js
+++ b/app/routes/ember-cli.js
@@ -1,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  titleToken() {
+    return 'Ember CLI';
+  }
+});
--- a/app/routes/ember-cli.js
+++ b/app/routes/ember-cli.js
@@ -1,7 +1,7 @@
 import Route from '@ember/routing/route';
 
 export default Route.extend({
-  titleToken() {
-    return 'Ember CLI';
+  title() {
+    return 'Ember CLI - Ember API Documentation';
   }
 });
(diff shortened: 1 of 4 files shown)


sivakumar-kailasam:
Looks like the review app deployment failed. I'll have a look at this PR once locally & get this sorted out.

app/routes/ember-cli.js
```diff
@@ -0,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  titleToken() {
+    return 'Ember CLI'
+  }
+})
```
mehulkar: @sivakumar-kailasam done. I didn't have time to look into how `titleToken` works, but with this change: `document.title` is "Ember - Ember CLI - Ember API Documentation". I tried empty string and not returning anything, but both of those remove the title entirely. In the other pages, that middle segment is the version, so there isnt' anything more appropriate possible here IMO. Let me know if you want a different string.

app/routes/ember-cli.js
```diff
@@ -0,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default Route.extend({
+  titleToken() {
+    return 'Ember CLI';
+  }
```
mehulkar: @sivakumar-kailasam done. I didn't have time to look into how `titleToken` works, but with this change: `document.title` is "Ember - Ember CLI - Ember API Documentation". I tried empty string and not returning anything, but both of those remove the title entirely. In the other pages, that middle segment is the version, so there isnt' anything more appropriate possible here IMO. Let me know if you want a different string.

sivakumar-kailasam: I believe the `title` function should work here. Made that change in this PR. Will look at the review app once to be sure.

PR #681, opened 2020-03-28
fix async request leaks in fastboot

Gaurav0:
Getting version ids should not require fetching all versions asynchronously. This prevents that and incidentally prevents async request leaks in fastboot, which I was originally trying to fix.

Thanks to @kiwiupover for all the help today!

### DIFF
--- a/tests/unit/utils/get-full-version-test.js
+++ b/tests/unit/utils/get-full-version-test.js
@@ -1,9 +1,10 @@
 import getFullVersion from 'ember-api-docs/utils/get-full-version';
 import { module, test } from 'qunit';
 import Service from '@ember/service';
-import Object from '@ember/object';
-import ArrayProxy from '@ember/array/proxy';
 import { A } from '@ember/array';
+import DS from 'ember-data';
+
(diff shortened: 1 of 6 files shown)


tests/unit/utils/get-full-version-test.js
```diff
@@ -14,11 +15,19 @@ module('Unit | Utility | get full version', function() {
         return '2.17.1';
       }
     });
-    let projectVersions = ArrayProxy.create({
-      content: A([{ id: 'ember-2.17.1'}, { id: 'ember-2.16.1' }])
+    let projectVersions = ManyArray.create({
```
sivakumar-kailasam: Could you please add some doc here on why `ManyArray` is needed here?

kiwiupover:
@Gaurav0 good work mate. 

This should also make the site faster too.

We should not be downloading every ember version to display the dropdown of ember versions.

tests/unit/utils/get-full-version-test.js
Gaurav0: @sivakumar-kailasam done.

PR #685, opened 2020-03-29
fix ember-native-dom-helpers deprecations

Gaurav0:


config/deprecation-workflow.js
```diff
@@ -3,6 +3,6 @@ self.deprecationWorkflow.config = {
   workflow: [
     { handler: "silence", matchId: "transition-state" },
     { handler: "throw", matchId: "deprecate-router-events" },
-    { handler: "silence", matchId: "ember-native-dom-helpers-click" }
+    { handler: "throw", matchId: "ember-native-dom-helpers-click" }
```
sivakumar-kailasam: Now that we've removed the addon, we can probably remove this config.

PR #694, opened 2020-04-03
fix computed override deprecation warnings

Gaurav0:


config/deprecation-workflow.js
```diff
@@ -1,6 +1,6 @@
 self.deprecationWorkflow = self.deprecationWorkflow || {};
 self.deprecationWorkflow.config = {
   workflow: [
-    { handler: "silence", matchId: "computed-property.override" }
+    { handler: "throw", matchId: "computed-property.override" }
```
sivakumar-kailasam, line 4: wouldn't this be the same as not having this configured here?

Gaurav0: Actually, no. The same would be a setting of "log"

Gaurav0:
This is now already included in master.

PR #696, opened 2020-04-04
fix fastboot

Gaurav0:


package.json
```diff
@@ -60,7 +61,7 @@
     "ember-composable-helpers": "^3.1.1",
     "ember-concurrency": "^1.1.6",
     "ember-data": "~3.16.0",
-    "ember-data-fastboot": "^0.1.2",
+    "ember-data-fastboot": "gaurav0/ember-data-fastboot#a681c9f3d3e550cc97520abd88272691170e4094",
```
kiwiupover, line 64: @Gaurav0 I would switch this to https://github.com/embermap/ember-data-storefront it does a better job with the shoebox

Gaurav0: Thanks! Will try it.

PR #699, opened 2020-04-05
fix meta store issues

Gaurav0:


Gaurav0:
Not working correctly anyway.

PR #719, opened 2020-05-07
set yarn policy

Gaurav0:
Ran `yarn set version 1.22.4` and committed the result.


Gaurav0:
It's not recommended to use one version of yarn to install another.

sivakumar-kailasam:
https://classic.yarnpkg.com/en/docs/package-json#toc-engines I was referring to this

Gaurav0:
I think that will just cause yarn install to fail if it doesn't match?

sivakumar-kailasam:
That’s true but Heroku uses that field to determine which version of yarn to use.

PR #742, opened 2021-02-11
Disable ligatures in code blocks

wagenet:
Before:
image
After:
image

See the `fi` in `fish` and `firstObject`.

### DIFF
--- a/app/styles/components/_highlight.scss
+++ b/app/styles/components/_highlight.scss
@@ -8,6 +8,7 @@
   margin: $small-spacing 0 $large-spacing;
   overflow: hidden;
   position: relative;
+  font-variant-ligatures: none;
 
   &.handlebars {
     .inline,


wagenet:
We may want to do this in some other repos as well.

jenweber (approved):
Thank you! Possibly our other apps are ok since they use different syntax highlighter libraries, if I remember right.

wagenet:
@jenweber what's the protocol for merging?

locks:
Boom, merged!

jenweber:
Thanks for the fix @wagenet! I was a little worried about the failed deploy preview. It has nothing to do with your PR, just maybe something that popped up elsewhere.

@locks did you run a deploy to production? Instructions are in DEPLOYING.md. It's been a while since I've done it.

PR #761, opened 2021-05-22
ember-data landing page: use ember-data-overview module main if present

runspired:
Results of pairing with @jenweber 

This allows the ember-data team to maintain and update the overview for ember-data directly with each release, falling back to a built-in overview for versions for which the library did not yet provide one.

I've configured it to use the main from the module `ember-data-overview` as opposed to `ember-data` to ensure that any old releases that had `ember-data` as a module (current releases do not) won't accidentally trip into this behavior.

We should consider expanding this for the ember-cli and ember.js teams to utilize as well, it would be quite easy to do.

The `ember-data-overview` page is filtered from the modules that will appear in the sidebar to prevent confusion, since it is already the primary link for the project. This also allows us to re-use `ember-data` module main as an explainer for that particular package vs for the project as a whole.

### DIFF
--- a/app/components/ember-data-landing-page.hbs
+++ b/app/components/ember-data-landing-page.hbs
@@ -1,4 +1,4 @@
-<article class="chapter">
+<<article class="chapter">
   <h1>
     Ember Data API Documentation
   </h1>
--- a/app/components/ember-landing-page.hbs
+++ b/app/components/ember-landing-page.hbs
@@ -1,0 +1,39 @@
+<article class="chapter">
+  <h1>Ember API Documentation</h1>
+  <p>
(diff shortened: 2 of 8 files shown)


jenweber:
@amyrlam  I’m going to set this PR up on the staging env and will let you know when it’s ready! Let me know if you think I should tag someone else for review.

amyrlam (approved):
:shipit: thank you for getting this across the finish line!

amyrlam:
@jenweber Can't think of anyone else to review, but I will xpost in #core-learning for awareness!

jenweber:
I click tested this in staging, and everything looks good there! Github has some pipeline issues affecting heroku, so this will be deployed after those clear up.

PR #764, opened 2021-06-03
WIP Add Lint to the Future

jenweber:
This tool allows us to progressively resolve linting errors and track our progress. Read about it here: https://www.npmjs.com/package/lint-to-the-future

Example in action: https://ember-learn.github.io/ember-styleguide/

It will need to have a GitHub workflow added in a future PR for us to view the dashboard. Example: https://github.com/ember-learn/ember-styleguide/blob/master/.github/workflows/gh-pages.yml
GitHub

Paired with @mansona


mansona:
Looks like we already added this in a different PR 🤔 I can't remember when we did it 🙃 https://ember-learn.github.io/ember-api-docs/

PR #766, opened 2021-06-11
remove page title

prakashchoudhary07:
Removed page-title, as it was already being updated


jenweber (approved):
Thank you for figuring this out!

Issue #775, opened 2021-10-01
Fix eslint issues 


rajakvk:
Fix eslint issues in the files merged along with PR #774 

Files:
1. app/adapters/application.js #776 
2. app/components/api-index-filter.js #777 
3. app/components/api-index.js #777 
4. app/components/class-field-description.js  #778
5. app/components/ember-anchor.js #779 
6. app/components/search-input/dropdown.js #781 
7. app/controllers/method.js #782
8. app/controllers/project-version.js #783
9. app/services/algolia.js #784
10. tests/unit/services/head-data-test.js #785

PR #776, opened 2021-10-01
Fix linting error (#775)

rajakvk:
Closes item-1 at #775


jenweber:
Thank you so much for your patience waiting for reviews. I will try to do better from here on out.

PR #777, opened 2021-10-15
Fix linting error (#775)

rajakvk:
Closes item-1 at #775


locks (approved):
Woohoo! Thanks for your effort 😁

PR #780, opened 2021-10-27
Display object properties

### LINKED ISSUE #765: Documented object properties are not displayed
One example is the [`deprecate` method](https://api.emberjs.com/ember/3.26/functions/@ember%2Fdebug/deprecate). I often forget what the options are and which of them are required. The current docs don’t show them, but they are [documented in the source](https://github.com/emberjs/ember.js/blob/v3.26.1/packages/%40ember/debug/lib/deprecate.ts#L177). I'm not sure if there are any more examples like this.

### Proposal

I propose we display these props. An indent should be enough to differentiate them from the function arguments.

#### Before

before

#### After

after

sdahlbac:
image

Not sure if this is the best way to style it, but seems to work at least.

Closes #765

### DIFF
--- a/app/styles/components/_class-field-desc.scss
+++ b/app/styles/components/_class-field-desc.scss
@@ -50,6 +50,10 @@ h3 .access{
   dd {
     display: inline-block;
   }
+  dl.parameters {
+    margin-left: 24px;
+    margin-block-start: 0;
+  }
 }
 
 .method, .property, .event {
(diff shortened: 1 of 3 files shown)


jenweber (approved):
Sorry this PR got lost in the notifications! This looks like a great improvement. I'll merge it if tests are still passing.

jenweber:
Thanks for your work! I click tested it locally and it's great to have this! Closes #797

Issue #788, opened 2022-01-07
Algolia search is broken


ctjhoa:
**Reproduction steps**
- Go to https://api.emberjs.com/ember/release
- Type any existing method in the search bar (eg. `factoryFor`)
- Note that suggested links are weird (eg. `https://api.emberjs.com/ember/release/classes/ember/methods/4.1?anchor=factoryFor`; suppose to be `https://api.emberjs.com/ember/4.1/classes/ApplicationInstance/methods/factoryFor?anchor=factoryFor`)
- Click on any suggested links
- You are redirected to https://api.emberjs.com/404

Browsers: Firefox 94.0.2, Chromium 96.0.4664.45

jenweber:
Thank you so much for reporting this! We are looking into it.

jenweber:
Search should be working in production now. I deployed a much earlier commit (`jw-search-debugging-2`) to get search working and buy us some time to find the bug.

I discovered that this is a front end problem in the ember app itself. It sure has the look of a mistake in some logic, where we are putting the version number where the path to the method should be.  This app went through a ton of revisions recently and it seems we missed something.

jenweber:
@mansona is going to have a look at this.

Examples:

Broken: http://localhost:4200/ember/release/classes/ember/methods/4.1?anchor=forEach
Correct https://api.emberjs.com/ember/4.1/classes/EmberArray/methods/forEach?anchor=forEach

Acceptance test we want to write:

- start at homepage
- does a search
- makes sure it goes to the right place

PR #791, opened 2022-01-22
Fix search

mansona:
Fixes https://github.com/ember-learn/ember-api-docs/issues/788

The issue was introduced automatically by a codemod in https://github.com/ember-learn/ember-api-docs/commit/d25f6b26d9ec94f45ba3ea956d093f5c521a0ce7 and unfortunately it looks like some of the other `{{link-to}}` modded code in that same commit has the same problem 🙈 

We should probably go through each one and make sure it works or at least is covered by a test and verify that it's broken

### DIFF
--- a/app/templates/components/search-input.hbs
+++ b/app/templates/components/search-input.hbs
@@ -6,6 +6,7 @@
   onfocus={{action 'onfocus'}}
   onblur={{action 'onblur'}}
   placeholder="Search"
+  data-test-search-input
   >
 {{!-- Search results dropdown --}}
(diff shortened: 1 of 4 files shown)


MelSumner (approved):
This seems correct but I'm not sure what's going on in Percy. Thank you for figuring this out!

mansona:
@MelSumner the issue with Percy is that we haven't merged anything since 4.0 and the version in the left hand side is out of date, and there seems to have been a class removed in 4.0 "Checkbox" 🤷 so all of the ToC have moved up one

locks (approved):

PR #792, opened 2022-01-28
Convert some more curly bracket links

jenweber:


chriskrycho (approved):

chriskrycho:
I *believe* CI should go green here now that I landed #819.

chriskrycho (approved):
Found and fixed the issue: needed `@models=` instead of `@model=` in one of these. CI should go ✅ now!

Issue #805, opened 2022-07-06
Resolve ember/no-classic-components deprecation


geneukum:
Right now, there are a bunch of classic components in ember-api-docs. It would be nice to work through those and upgrade them all to use native syntax and extend from glimmer/component. The components requiring modifications are:

- [ ]  app/components/api-index-filter.js
- [ ]  app/components/api-index.js 
- [ ]  app/components/class-field-description.js
- [ ]  app/components/import-example.js
- [ ]  app/components/loading-spinner.js 
- [ ]  app/components/search-input.js 
- [ ]  app/components/search-input/dropdown-header.js
- [ ]  app/components/search-input/dropdown-result.js 
- [ ]  app/components/search-input/dropdown.js
- [ ]  app/components/table-of-contents.js

Once we've done this we should then:

- [ ]  re-enable the ESLint rule: 'ember/no-classic-components' 

to prevent us from accidentally adding any legacy style components going forward.

PR #809, opened 2022-07-20
Update table of contents component to glimmer

geneukum:
Let's update the TableOfContents component to extend glimmer/component instead of ember/component. Glimmer components offer a bunch of various advantages over classic components and are the direction that ember is
heading in general.

While we're here, let's prefer single quotes to double quotes throughout our template file and cleanup any unused arguments into this component. 

This should resolve the `app/components/table-of-contents.js` section of https://github.com/ember-learn/ember-api-docs/issues/805.

### DIFF
--- a/app/components/table-of-contents.hbs
+++ b/app/components/table-of-contents.hbs
@@ -1,12 +1,12 @@
-<ol class="toc-level-0">
-  <li class="toc-level-0">
-    <a {{action 'toggle' 'modules'}} href="#" data-test-toc-title="packages">Packages</a>
-    <ol class="toc-level-1 modules selected">
+<ol class='toc-level-0'>
+  <li class='toc-level-0'>
+    <a {{action 'toggle' 'modules'}} href='#' data-test-toc-title='packages'>Packages</a>
+    <ol class='toc-level-1 modules selected'>
       {{#each @moduleIDs as |moduleID|}}
 
-        {{#if (not-eq moduleID "@ember/object/computed")}}
-          <li class="toc-level-1" data-test-module={{moduleID}}>
-            <LinkTo @route="project-version.modules.module" @models={{array @version moduleID}}>{{moduleID}}</LinkTo>
+        {{#if (not-eq moduleID '@ember/object/computed')}}
+          <li class='toc-level-1' data-test-module={{moduleID}}>
+            <LinkTo @route='project-version.modules.module' @models={{array @version moduleID}}>{{moduleID}}</LinkTo>
           </li>
         {{/if}}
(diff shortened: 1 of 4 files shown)


tests/integration/components/table-of-contents-test.js
```diff
@@ -12,18 +12,16 @@ module('Integration | Component | table of contents', function (hooks) {
...
     await render(hbs`
-      {{
-        table-of-contents showPrivateClasses=true
-          projectid=projectId
-          version=emberVersion
-          classesIDs=classesIDs
-          isShowingNamespaces=true
-       }}
+      <TableOfContents
+        @showPrivateClasses={{true}}
```
chriskrycho, line 20: Note for a follow-on: this currently uses two-way binding between an `Input` in the implementation and a property on a service very far away. As a follow-up, we'll want to update that to actually just use normal event handling.

app/components/table-of-contents.hbs
```diff
@@ -0,0 +1,44 @@
...
+          <li class='toc-level-1' data-test-namespace={{namespaceID}}>
+            <LinkTo @route='project-version.namespaces.namespace' @models={{array @version namespaceID}}>{{namespaceID}}</LinkTo>
+          </li>
+        {{/each}}
+      </ol>
+    </li>
+  {{/if}}
+
+  <li class='toc-level-0'>
+    <a {{action 'toggle' 'classes'}} href='#' data-test-toc-title='classes'>Classes</a>
```
chriskrycho, line 31: It is *good* that these didn't change in this case, but we'll definitely want to switch to `{{on}}` in a follow-up.

app/components/table-of-contents.hbs
```diff
@@ -0,0 +1,44 @@
...
+      {{#each @classesIDs as |classID|}}
+        <li class='toc-level-1' data-test-class={{classID}}>
+          <LinkTo @route='project-version.classes.class' @models={{array @version classID}}>{{classID}}</LinkTo>
+        </li>
+      {{/each}}
+    </ol>
+  </li>
+</ol>
+<label class='toc-private-toggle'>
+  <Input @type='checkbox' @checked={{@showPrivateClasses}} class='private-deprecated-toggle' />
```
chriskrycho, line 42: And this is the two-way binding we'll want to replace.

chriskrycho (approved):

chriskrycho:
This specific CI failure is happening on `master`, which is why this was green before updating it with the latest changes from `master`, so I'm going to merge past it. I'll do a separate PR to fix it.

PR #812, opened 2022-07-22
Update SCSS division syntax

jenweber:
Using / for division was deprecated and will be removed in Dart Sass 2.
We can replace it with math.div.

Co-authored-by: Chris Manson <chris@manson.ie>


mansona (approved):
🎉

PR #815, opened 2022-08-05
Use prember

jenweber:
Project board: https://github.com/orgs/ember-learn/projects/68/views/1

We aim to use prember to build the docs site, eliminating the need for a back end server. We also aim to store docs json in git instead of S3, which means we can edit those docs by hand when needed, fix broken URLs, etc. and not have to run new releases to do so.

It is less work if we do this with git and not worry about shipping it for the old json docs. The reason is we can clone the git repo easily in order to get the data prember needs, instead of fetching a zillion things from s3

To update the local branch so that it contains the json docs, use `npm run clone`. Otherwise see an error about a directory not existing.


jaredgalanis (approved):
🚀

PR #826, opened 2022-11-19
Demixinify

chriskrycho:
- Inline filter-params mixin into its only user
- Replace parent-name mixin with simple getter and util


chriskrycho:
Not sure why Percy failed here, either. 🤔 Trying the "merge in `master` and see if that helps" dance.

chriskrycho:
All right, since the last commit here was the one which changed the FastBoot-related behavior, I've dropped that one and will try to take a further swing at it on a different PR; I also extracted what was accidentally the first commit here and extracted it to #830.

chriskrycho:
There we go. I'll tackle the `transitionTo` and the last mixin separately!

PR #827, opened 2022-11-19
Follow-up items from updating table of contents component to glimmer

geneukum:
Let's follow-up on the suggested tweaks to the table-of-contents component from: https://github.com/ember-learn/ember-api-docs/pull/809 by:

1.  avoiding the `action` helper 
2. avoiding the `Input` component which provides a 2-way binding on the `@checked` arg

### DIFF
--- a/app/controllers/project-version.js
+++ b/app/controllers/project-version.js
@@ -1,5 +1,5 @@
 /* eslint-disable ember/no-computed-properties-in-native-classes */
-import { computed } from '@ember/object';
+import { action, computed, set } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { readOnly, alias } from '@ember/object/computed';
 import Controller from '@ember/controller';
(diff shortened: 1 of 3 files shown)


app/controllers/project-version.js
```diff
@@ -121,4 +121,9 @@ export default class ProjectVersionController extends Controller {
 
   @readOnly('model.project.id')
   activeProject;
+
+  @action
+  togglePrivateClasses() {
+    set(this, 'showPrivateClasses', !this.showPrivateClasses);
+  }
```
chriskrycho, line 128: Let's also follow up on *this* and get rid of the CP macro in favor of actually using tracked state on the service. I know that the service and the way these CPs work are a bit janky today, *but* I did a spike where I just swapped it all out for a simple set of tracked properties and *not* trying to do those janky things in the CPs and it actually “just worked” (though we should of course verify that).

chriskrycho (approved):

PR #828, opened 2022-11-19
Update to use Yarn v3

chriskrycho:
While this project may ultimately switch back to using npm, or may opt to use pnpm instead, Yarn 3 is a significant win for as long as this app *is* using Yarn:

1. Yarn v3 is maintained; Yarn v1 is *not*. This is the most important reason to make this switch.

2. Yarn v2+ handles peer dependencies much more rigorously than Yarn v1 (or npm), albeit not quite as well as pnpm. As part of the ecosystem's move to using properly Node-resolvable packages, we are increasingly making use of peer dependencies, so this is important future-readiness.


jenweber:
We chatted about this very briefly at the learning team meeting! We need a little time to let everyone read and think about this. We will check in again next week!

PR #830, opened 2022-11-19
Switch away from deprecated `Route.transitionTo`

chriskrycho:
This sets us up nicely for Ember v4, where this isn't present. It also removes an unused mixin!


chriskrycho:
Odd, unclear why tests are failing. I will poke at this at some point Soon™; if someone else beats me to it, help welcome!

geneukum:
TIL - per [the RouterService docs](https://api.emberjs.com/ember/release/classes/RouterService/methods/transitionTo?anchor=transitionTo):

> Calling transitionTo from the Router service will cause default query parameter values to be included in the URL. This behavior is different from calling transitionTo on a route or transitionToRoute on a controller.

Adding this extra `?show=inherited` query string to the expected URL seemed to get all of those tests passing again.

chriskrycho:
Ah, yep, that would do it! Thank you!

PR #831, opened 2022-11-19
Add Glint to the project

chriskrycho:
This adds basic [Glint]-powered capabilities to the project for anyone using a Glint Language Server client (VS Code, NeoVim, IntelliJ, etc.).

Demo
----

https://user-images.githubusercontent.com/2403023/202874192-1dc3e6bf-8185-4ffc-8e40-e7d6d6dba825.mp4

Usage
-----

This won't impact you *at all* unless you have a Glint language server client set up. You can install the Glint extension for VS Code, the IntelliJ Ember extension, or the vim/NeoVim coc-glint plugin, and this will "just work"; if you don't use those it will stay out of your way!

Setup
-----

The setup to make that work involves adding some `devDependencies`:

- `@glint/core` to enable Glint and `@glint/environment-ember-loose` to tell Glint  how to interpret Ember's "loose-mode" templates, i.e. "normal"

- `typescript`, which Glint requires to work at all. **This is _not_ adding TypeScript to the project.** It just exists to enable Glint to run. (Glint doesn't ship its own copy because TypeScript is large and it is important that Glint use the same version of TypeScript that an app uses, or else it would give conflicting reports.)

- A number of `@types` packages, which are required so that Glint knows what a `Controller` or Ember `Component` are, and so on. These will all be able to go away as packages start shipping their own types, including when this app is able to upgrade to at least Ember 4.8 and start using the types that Ember ships itself.

It also configures Glint via tweaks to the existing `jsconfig.json` file, with good defaults so that it never shows *errors*, but does progressively enhance the experience for JS users.

Other changes
-------------

With the foundations in place, add some JSDoc comments and a single TS file to the project.

- The JSDoc comments show how to make TS-powered tooling (like the various Glint language servers) understand components in the app. These comments, which build on the spec for both types and comments defined in [Ember RFC 0748][rfc], work well to enhance the experience of authoring JS backing classes even *without* Glint. With Glint enabled, though, it also makes all of that information available in templates.

- The TS file shows how to make Glint understand component references. It is basically a hand-coded version of what the Ember resolver does at runtime: translating string names like `MyComponent` into references to the `MyComponent` class. There are limitations to how much this can do in a purely JS project with loose mode templates, but in this app it can cover *most* of the territory (and a move to strict mode templates using `<template>` in the future will close *all* of those gaps).

Both of these show the "progressive enhancement" model at work: out of the box, you get a *very* minimal set of go-to-definition, refactoring, etc. capabilities. Not none! But minimal. The more info you put into your JSDoc comments, though, and the more items you add to the template registry, the more useful Glint becomes.

[rfc]: https://rfcs.emberjs.com/id/0748-glimmer-component-signature

Caveats
-------

This does not yet provide *all* the benefits that Glint can provide to JS-only apps. For example, you won't be able to go-to-definition for the `this.model` of route/controller templates unless you add a `model` field on the `Controller` class with a JSDoc annotation. Similarly, you will not be able to go-to-definition or see hover docs for a lot of the data used in the application just yet: the app will need to add some JSDoc annotations to tell the language server "this field here is that data model class from over there".

However, it *does* already provide significant benefits for simple things like being able to go to definition or rename items in component files, and as the one JSDoc annotation added in this shows, you can get more of these benefits with a very low level of effort.

[Glint]: https://github.com/typed-ember/glint


jenweber:
We chatted about this very briefly at the learning team meeting! We need a little time to let everyone read and think about this. We will check in again next week!

jenweber:
Now that I’ve read through the details - I’m happy to add an opt-in developer experience enhancement. I say we try it out. What is maintenance like, apart from bumping package versions periodically?

I still want to give some time for others to read before we merge.

chriskrycho:
Bumping versions occasionally should be basically it. Keeping docs up-to-date for what things are internally will also be valuable, though it won't break anything if that is wrong, except regressing developer experience. But that's the same as the story for JS today!

jenweber:
Ok, thanks for the info!

MinThaMie:
To support the large refactor you and the gang are undertaking we approve this PR. We do not want to enforce types in CI in order to not block developers who don't do typescript.

Is it truely okay to completely ignore the types when Chris Manson and Anne-Greeth start working on the StyleGuide stuff for this project?

MinThaMie (approved):

PR #834, opened 2022-11-21
Get the main branch deployable by restricting node versions

jenweber:
I found that our deployment pipeline does not read from volta. If we restrict to Node 14 in `engines` of package.json, the app deploys 🎉 

Alternate options:

- test if Node 16 works (I haven't done this yet)
- could use npmrc
- try to fix what broke when using Node 18 (I think we should do this, just maybe not _now_ given other priorities)

I have a slight preference for using engines, in the interest of not proliferating places that version compat could live. Open to other approaches. 

QA click testing available at https://ember-api-docs-staging.herokuapp.com/ember/release


chriskrycho:
The issue with Heroku is annoying; notably, though, engines is much less reliable than Volta and, critically, you cannot use Volta to specify a package manager unless it specifies Node. Net, there's a tradeoff here in terms of having a single source of truth vs, how much that source of truth actually gives you. This change certainly makes sense to unblock, regardless, and the Learning team can evaluate the Volta question separately that way. (Has anyone checked into whether Heroku does, or has any plans to, support using Volta?)

chriskrycho (approved):

jenweber:
I think we’re stuck with engines. I googled “Volta heroku support” and came up with nothing. Heroku docs talk about engines only: https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

Let’s keep both Volta and engines.

PR #835, opened 2022-11-21
Use NPM instead of Yarn

MinThaMie:


chriskrycho:
Seems fine to me as an outside contributor: I don't especially care. However, there are a couple things to consider when evaluating this as an alternative to either upgrading to Yarn 3 or switching to pnpm:

- While npm's performance has improved, such that it is now comparable to Yarn 1, it is still much, *much* slower than Yarn 3 or pnpm. That is not a make or break thing, but it is a really nice win as a contributor.

- Unfortunately, npm still does not handle `peerDependencies` very well, while both Yarn 3 and pnpm have significantly improved in that area. We are steadily working (especially as part of Embroider) to make `ember-source` and other core packages be well-behaved Node packages in terms of how they and libraries which use them get folded together by standard build tools, and that means a lot more uses of `peerDependencies` (roughly: there should only be one actual copy of Ember itself in the thing you build!).

Net, while npm is *fine*, and has the not insignificant upside of being bundled with Node, I am encouraging folks in general across the community (not just in this app!) to look at using either Yarn 3 or pnpm, both of which are also increasingly easy to use (and provide good DX to users out of the box) courtesy of corepack.

MinThaMie:
Thanks for you elaborate explanation :) Let's see what the rest things of this. We discussed it yesterday in the learning-meeting and would prefer to move to NPM first and who knows when we would move to pnpm

mansona (approved):
Thanks for doing this @MinThaMie 🎉 very happy to see this

PR #848, opened 2022-12-21
Colocate remaining components

chriskrycho:
This is good prep for landing `<template>` (colocated templates have the exact same semantics for runtime code as `<template>` does) , and a mild performance improvement to boot since we have fewer modules to manage at runtime.


bertdeblock (approved):

mansona (approved):
looks good 👍

PR #851, opened 2023-01-27
Bugfix for Ember Data landing page

jenweber:
Older versions of Ember Data were unreachable if they did not have an "overview" module.

Reproduction:
- visit api.emberjs.com
- select a version older than 4.6 from the dropdown
- get a 404

This is not the nicest way to fix this, but I think it works.


runspired (approved):

mansona (approved):
looks good to me 👍

PR #861, opened 2023-04-28
skip tests that are broken because of invalid data in production

mansona:
Our tests currently rely on downloading data from the real API (a problem but something we have a plan for 🙈)  but there was an issue with a recent deployment that caused some URLs to fail

e.g. https://api.emberjs.com/ember/3.13/functions/@ember%2Fapplication/getOwner works but https://api.emberjs.com/ember/3.12/functions/@ember%2Fapplication/getOwner does not

To unblock all PRs to this repo I'm just skipping tests that rely on that data and duplicating them to use more modern versions that aren't broken 👍 we can un-skip the old tests once the issue with the production data is fixed

### DIFF
--- a/tests/acceptance/function-test.js
+++ b/tests/acceptance/function-test.js
@@ -5,8 +5,19 @@ import { visit, find } from '@ember/test-helpers';
 module('Acceptance | Function', function (hooks) {
   setupApplicationTest(hooks);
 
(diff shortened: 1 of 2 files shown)


jaredgalanis (approved):
LGTM 👍

PR #862, opened 2023-05-01
Bugfix - fix broken ember-data GitHub URLS

jenweber:
If the file has packages already in the path, make sure we don't add duplicate packages via the mainDir function.
Fixes an issue with ember data URLS having an incorrect 'packages/packages' in the GitHub source URL

Example problem was
https://api.emberjs.com/ember-data/4.10/classes/RecordArray

Closes ember.js  #20404

### DIFF
--- a/app/helpers/github-link.js
+++ b/app/helpers/github-link.js
@@ -1,6 +1,12 @@
 import { helper } from '@ember/component/helper';
 import githubMap, { mainDir } from '../utils/github-map';
 
+/**
+ * https://github.com/emberjs/data/tree/v4.10.0/packages/packages/
+ *   store/addon/-private/record-arrays/identifier-array.ts#L118
+ *
+ * File: ../packages/store/addon/-private/record-arrays/identifier-array.ts
+ */
 export function githubLink([project, version, file, line], { isEdit = false }) {
   if (isEdit) {
     return `https://github.com/${githubMap[project]}/edit/release${mainDir(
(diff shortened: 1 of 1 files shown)


jenweber:
@runspired do you have any idea why there were recently some duplicate `packages/packages` in the links that go to GitHub?  This PR above patches it on the front end, but maybe there is something to fix in the ember-data source code too so that this workaround doesn't need to be active for versions moving forward. Let me know if you have any clues!

jaredgalanis (approved):
LGTM! 👍

app/helpers/github-link.js
```diff
@@ -1,17 +1,30 @@
...
-  return `https://github.com/${githubMap[project]}/tree/v${version}${mainDir(
-    project,
-    version
-  )}${file}#L${line}`;
+
+  // If the file has packages already in the path, make sure we don't
+  // add duplicate packages via the mainDir function.
+  // Fixes an issue with ember data URLS having an incorrect
+  // 'packages/packages' in the GitHub source URL
+  const fixedFile = file.replace('../packages/', '../');
```
jaredgalanis, line 22: So I approved this, but then saw that `file` can be undefined potentially (some tests are failing with an error thrown in that way). I can take a look after working hours.

jaredgalanis (changes requested):
we need to address the failing tests b/f merging this

jaredgalanis (approved):

runspired:
This is fixed in EmberData now https://github.com/emberjs/data/pull/8598

jenweber:
Awesome, thank you!

Issue #865, opened 2023-05-27
Blank page error after ember 3.28 update


CvX:
There's an error `undefined is not an object (evaluating 'e.map')` (in is-latest/getLastVersion but stems from a ember-data relationship issue) and effectively a blank page when visiting e.g. https://api.emberjs.com/ember/3.28/classes/Ember.Templates.components/methods/LinkTo?anchor=LinkTo

This regressed in #860 (cc: @mansona)

bearcott:
should be fixed

Issue #801, opened 2022-04-20
Page not found in API documentation


lfekson:
"ACK! 404 FRIEND, YOU'RE IN THE WRONG PLACE" when you click on https://api.emberjs.com/ember/release/classes/TextField link on https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input?anchor=Input page.

jacobq:
Thanks for reporting. When I've run into stuff like this before, annoying as it is, I've often found that it's due to deprecated or removed API (i.e. the thing being referred to doesn't exist in the version you're looking at, but the text/links haven't been updated to reflect this). I do hope that someone will fix the docs site generator code to better support historical versions (e.g. when viewing the v3.28 docs I don't want links to jump to release/4.x version). In the meantime, however, once you know about this, you can try replacing `release` with `3.28` (or whatever version you want) in the URL.

In this case, built-in components [were removed in 4.x](https://deprecations.emberjs.com/v3.x#toc_ember-built-in-components-import). If you want to continue using them you have to add them back via a legacy addon. You can see the docs for the 3.28 version, which still included that deprecated feature here: https://api.emberjs.com/ember/3.28/classes/TextField

PR #866, opened 2023-05-31
bugfix: hot-fix for blank pages on some routes

bearcott:
as noted in #865 you get a blank page when visiting pages like:
- https://api.emberjs.com/ember/3.28/classes/Ember.Templates.components/methods/LinkTo?anchor=LinkTo
- http://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input

The immediate cause for the bug is the "Edit on Github" button breaking the page:
Screenshot 2023-05-30 at 7 06 06 PM

From my findings something is causing these page to render twice, the first time `model.project.id` is `undefined` for some reason. This causes the latest version check in the conditional to break. Imo avoiding this initial render would be the optimal fix so if anyone has an idea whats causing it that would be good!

For now I suggest a hot fix to check if `model.project.id` exists before rendering the component as well as `model.file` and `model.line` since they are also undefined for some reason. The component shouldn't be rendered anyways if any of these are undefined since it will just be a dead link.

### DIFF
--- a/app/templates/project-version/classes/class.hbs
+++ b/app/templates/project-version/classes/class.hbs
@@ -1,6 +1,6 @@
 {{! template-lint-disable no-action }}
 <article class="chapter">
-  {{#if (is-latest version=@model.projectVersion.version allVersions=this.allVersions)}}
+  {{#if (and @model.project.id @model.file @model.line (is-latest version=@model.projectVersion.version allVersions=this.allVersions))}}
     <a data-tooltip="Edit on Github" class="heading__link__edit" href="{{github-link @model.project.id @model.projectVersion.version @model.file @model.line isEdit=true}}" target="_blank" rel="noopener noreferrer">{{svg-jar "fa-pencil"}}</a>
   {{/if}}
   <h1 class="module-name">Class {{@model.name}}</h1>


jenweber:
Thank you for the hotfix! That is very odd. We recently did some version updates and I’m guessing a bug got introduced. I’ll do a click test and deploy.

jenweber:
@mansona I am wondering if we broke something at the 3.28 upgrade? Or possibly this was an earlier issue.

jenweber:
Actually I'll have to wait a day or so to deploy. Heroku had some service problems today, and things still seem shaky. If any fellow learning team members want to merge and deploy to staging and then prod, after the Heroku issue clears up, that would be great. Instructions are in MAINTAINERS.md. I have click tested locally and this is good to go.

> The Heroku Platform API is unavailable. Customers using the API, the CLI, and the Heroku Dashboard will be unable to make changes/releases to or queries of their applications. Applications currently running will continue to run. Engineers are investigating, we'll update as we have more details.

 https://status.heroku.com/

bearcott:
@jenweber thanks for the context. May I suggest possibly it's an API side issue? When I did some digging both the `line` and `file` references were undefined for `Input` I'm guessing [may be related to deprecated builtins](https://github.com/ember-learn/ember-api-docs/issues/801).

jenweber:
@bearcott yeah I think you are on the right track, thinking that this probably has roots in the data. Unfortunately right now, the only way to change the underlying docs data is through running new releases of the libraries themselves (like ember-data) but we are working on a fix that will let us change things after the fact! So I will try to remember this case and look for it.

jenweber (approved):

Issue #872, opened 2023-07-24
Stale search results when switching versions


a13o:
**Steps to reproduce**

1. Set the version dropdown; e.g. `5.1`
2. Search for a term in the global search at the top; e.g. `get`
3. Set the version dropdown to something else; e.g. `4.12`
4. Click on the global search at the top again

**Expected behavior**

Search results are relevant to the most recently selected version, `4.12`

**Actual behavior**

Search results are stale and reference the originally selected version, `5.1`

PR #873, opened 2023-07-24
Clear stale search results when version is changed

a13o:
Possible fix for #872. This approach replaces the stale search results just in time whenever the search results dropdown is about to display.

By doing it this way we don't have to clear the user's search term whenever they change versions. They can pick up right where they left off, but with fresh search results.

todo
- [ ] needs tests
- [ ] clearing the results before calling `search.perform()` may eliminate the flash of stale content inside the dropdown
- [ ] what happens if they change projects, but leave version alone?

### DIFF
--- a/app/components/search-input.js
+++ b/app/components/search-input.js
@@ -56,6 +56,10 @@ export default class SearchInput extends Component {
   }
 
   @action onfocus() {
+    if (this.query.length > 0 && this.searchService.hasStaleResults()) {
+      this.searchService.search.perform(this.query);
+    }
+
     this._focused = true;
   }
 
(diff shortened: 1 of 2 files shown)


jenweber:
Thanks for the fix! Is this ready for review?

I suspect that maybe this feature got a bug when we converted some components to use tracked instead of observers and computed properties. It will be great to have it working!

a13o:
Oops, I have some lint issues to fix, but the general approach is ready for feedback

MinThaMie (approved):
This looks great :) Thanks for fixing this and for the elaborate demo video!

locks (approved):

Issue #879, opened 2023-08-24
Start using "wide mode" in the redesign branch


mansona:

mansona:
@lukasnys is going to take a look at this one too 👍

PR #881, opened 2023-09-02
API docs styling tweaks

lukasnys:
Fixes https://github.com/ember-learn/ember-api-docs/issues/879 as well as some other styling tweaks:
- Use `sidebar-container--full-width` (#879) and add necessary padding
- Match sidebar font-size with guidemaker-ember-template (PR: https://github.com/ember-learn/guidemaker-ember-template/pull/162)
- Make header full-width
- Increase sidebar width so items don't overflow

Let me know if one of the additional styling tweaks (font-size changes, header width changes, sidebar width changes) are undesired as I didn't talk this through with anyone.

<details>
<summary>Screenshots before and after</summary>

| x | Before | After |
|--------|--------|--------|
| **Desktop** | localhost_4200_ember_release (2)   | localhost_4200_ember_release (1) |
| **Mobile** | localhost_4200_ember_release_modules_@ember%2Farray (2) | localhost_4200_ember_release_modules_@ember%2Farray (3) |
| **Mobile sidebar** | localhost_4200_ember_release_modules_@ember%2Farray (4)  | localhost_4200_ember_release_modules_@ember%2Farray (5)  |
</details>

### DIFF
--- a/app/styles/app.css
+++ b/app/styles/app.css
@@ -26,6 +26,11 @@ main a.edit-icon {
   background: none;
 }
 
+.es-header {
+  padding: 0 var(--spacing-4);
+  justify-content: start;
+}
+
 .es-sidebar-content .table-of-contents:first-child {
   margin-block-start: 0;
 }
--- a/app/styles/app.css
+++ b/app/styles/app.css
@@ -26,6 +26,21 @@ main a.edit-icon {
   background: none;
 }
 
+@media (min-width: 845px) {
+  .es-sidebar {
+    padding: var(--spacing-4);
+    background-color: var(--color-gray-200);
+  }
+}
+
+.es-sidebar-content .table-of-contents:first-child {
+  margin-block-start: 0;
+}
+
+.content {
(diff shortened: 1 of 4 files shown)


lukasnys:
One thing I'm concerned with in this PR is the `--sidebar-width: min-content`. I added it to stop sidebar items from overflowing their parent, however if the sidebar items ever become smaller in size, it might cause the layout to "break"

MinThaMie:
Am I correct when I understand that you made it `min-content` because the sidebar-items were not wrapping? Because I think we would love it if the sidebar-width is consistent between all the "apps"/websites and using min-content does not guarantee that.

MinThaMie (commented):
Am I correct when I understand that you made it min-content because the sidebar-items were not wrapping? Because I think we would love it if the sidebar-width is consistent between all the "apps"/websites and using min-content does not guarantee that. So if that's what you are aiming for, I suggest using word-wrap.

MinThaMie (approved):

PR #900, opened 2023-12-08
update node-sass and ember-auto-import

mansona:
This should allow us to finally run this on modern node 🎉


jaredgalanis (approved):
LGTM 👍

PR #909, opened 2024-06-14
Removing the tabbed-layout and moving it to an on-this-page layout

MinThaMie:
Screenshot 2024-06-14 at 09 49 42


MinThaMie:
We'll move the sticky and updated styling that is currently in the guides to style guide and will update style guide in the website-redesign branch. That will make the on-this-page behave nicer and look better :)

PR #923, opened 2024-07-20
stop requesting mappings over the network and fix shoebox usage

mansona:
for some reason we're requesting the mappings.json file on every initial page load (and not cached in the shoebox). This is in the application afterModel hook so there is no chance that we will ever not make this request, so it's worth just backing the json into the JS bundle to save a request


MinThaMie (approved):
LGTM :)

PR #934, opened 2025-01-28
fix: update GitHub Pages workflow to use Node.js 20 and pnpm 9

MehulKChaudhari:
This PR updates the GitHub Pages workflow to use Node.js 20 and pnpm 9, matching the local setup and CI.yaml to fix deployment failures.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -10,12 +10,13 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
         with:
-          node-version: '14'
-      - run: npm install -g npm@7
-      - run: npm install
+          node-version: '20'
+      - run: npm install -g pnpm@9
+      - run: pnpm install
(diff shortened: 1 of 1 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -10,12 +10,13 @@ jobs:
...
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
         with:
-          node-version: '14'
-      - run: npm install -g npm@7
-      - run: npm install
+          node-version: '20'
+      - run: npm install -g pnpm@9
```
mansona, line 18: you shouldn't need to do this global install, you can use pnpm/setup-action which means that you can get a better cache implementation: 

```
- uses: pnpm/action-setup@v4
- uses: actions/setup-node@v4
    with:
      cache: 'pnpm'
      node-version: 20
- run: pnpm i --frozen-lockfile
```

mansona (changes requested):
Thanks for keeping this up to date 🎉 if we're updating it we should make sure that we're updating it to the best practice (which is kind of hard to see in the other github action because of the use of variables for node versions)

Also it's worth noting that this action is just deploying the lint-to-the-future dashboard on https://ember-learn.github.io/ember-api-docs/ so we don't need to wait for the full deployment to be sure that this will be ok to merge 👍

.github/workflows/gh-pages.yml
MehulKChaudhari: Yes, sounds good. Made the changes

mansona (approved):

PR #936, opened 2025-02-27
update ember-power-select and ember-concurrency

mansona:
I'm trying to use the GJS codemod and ember-power-select was a blocker 👍


MinThaMie (approved):
Thanks :)

PR #939, opened 2025-03-21
chore: remove node-sass

runspired:
I'm sure this doesn't work and is all wrong but a blind attempt at removing node-sass which has no builds that work with either a maintained node version or a maintained OSX version at this point.


kategengler:
Resolved by merging #868

PR #941, opened 2025-05-22
fix implicit-injections deprecation

mansona:


app/routes/project-version/modules/module.js
```diff
@@ -1,8 +1,11 @@
 import ClassRoute from '../classes/class';
 import ScrollTracker from 'ember-api-docs/mixins/scroll-tracker';
 import getFullVersion from 'ember-api-docs/utils/get-full-version';
+import { inject as service } from '@ember/service';
```
IgnaceMaes, line 4: I suppose importing service is not supported on this `ember-source` version yet?

IgnaceMaes (approved):

app/routes/project-version/modules/module.js
mansona: that's correct, we're still on 3.28 here 🙈

PR #956, opened 2025-08-25
Fixes to class item descriptions

kategengler:
- Make arguments and return types more easily distinguishable from the method names (and closer to typedoc)
- Fix missing argument to show the module 
- Fix for the UI when a class item wraps


MinThaMie (approved):
Looks great, thanks for the little design tweaks and fixing/finding a way to do the redirects

